PR #9, opened 2016-02-03
MysqlInternalError contains MySQL error code of occured error.

### LINKED ISSUE #8: QueryError should have error code.
QueryError should have separate member containing error code to be able to programatically decide what action to take when exception occurs.

Other option is to have separate exception class for each error code, but the first option is definitely easier and it should be enough.

niximor:
Fixes #8.


danopernis:
Merged (with minor tweaks)

PR #13, opened 2016-06-22
Add row stream

danopernis:


include/superior_mysqlpp/row_stream.hpp
```diff
@@ -0,0 +1,43 @@
...
+        explicit RowStream(Row& row)
+            : first {row.begin()}
+            , last {row.end()}
+        {
+        }
+
+        template<typename T>
+        RowStream& operator>>(T& obj)
+        {
+            assert (first != last);
```
majcl: No assert. Use exception.

include/superior_mysqlpp/row_stream.hpp
```diff
@@ -0,0 +1,43 @@
...
+            if (!(*first).isNull()) {
```
majcl: 1) What about Nullable? 
2) When value is null - nothing will change, user will not know, that nothing happened.

include/superior_mysqlpp/row_stream.hpp
danopernis: This adapter is intended for callers who are aware of how many columns there are in resulting row. I can provide `operator bool()` to check if the extraction is valid for callers who do no want to keep track of the number themselves. Assert just documents this precondition.

Exception is costly. Assert is noop when NDEBUG is #defined.

include/superior_mysqlpp/row_stream.hpp
danopernis: Nullable is a concept in prepared statements, I do not see how it applies here.

You might use `boost::optional<T>` to detect if the value is indeed present.

include/superior_mysqlpp/row_stream.hpp
majcl: Library should be safe, what is the cost, for not thrown exception here?

tests/db_access/row_stream.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <superior_mysqlpp.hpp>
+
+#include "settings.hpp"
+
+
+using namespace bandit;
+using namespace SuperiorMySqlpp;
+
+go_bandit([](){
+    describe("Test RowStream", [&](){
```
majcl: Make here test for all types please.

majcl:
Actually this is completely different use philosophy, than was intended to.
Also, problem is that with dynamicPreparedStatement we are almost in same situation(we know results in runtime), but it has completely different use style.
I understand someone can/want use this style. 
Please make there directory extras and put it there. Also please change its name to something Adapter or Helper or something...
Also please create special section in documentation with examples.

include/superior_mysqlpp/row_stream.hpp
majcl: Ok, provide operator bool there.

include/superior_mysqlpp/row_stream.hpp
majcl: If T is already set, then >> on T obj will do nothing. It will not unset optional. This is problem.

tests/db_access/row_stream.cpp
danopernis: Other types are already covered in tests for converters and types.

include/superior_mysqlpp/row_stream.hpp
danopernis: Good point, I will alter the behaviour to assign a default-constructed `T` in NULL case.

danopernis:
I have pushed some fixup commits and squashed them into new branch `add_row_stream-2`

PR #17, opened 2016-07-11
fix for value() method of nullable

FUSAKLA:
Fix of bug in method value() in nullable type.


FUSAKLA:
Found out bug is not in the library so fix won't be needed probably.

PR #18, opened 2016-07-25
fix engaged not set in nullable when using dynamic prepared statement

stanislav-raclavsky:


stanislav-raclavsky:
update after review -> new pull request

Issue #4, opened 2016-01-06
TCP port default in SuperiorMySqlpp::Connection() makes connecting to socket less comfortable.


jschrotter:
It's necessary to connect using unix socket using:
`SuperiorMySqlpp::Connection(database, user, password, socket, std::make_tuple())`
Removing default TCP port from this constructor solves it.

smradCZ:
It was solved by #22

PR #22, opened 2016-11-01
Added ConnectionConfig including tests

jirislav:
ConnectionConfig is a class which should solve the problem with ambiguity while calling Connection c'tor with only host / socket defined (compiler error)

Here is the default value for c'tor with socket unset for socket so that there is no more ambiguity and is added a ConnectionConfig class instead. It provides two static factory methods for both TCP Connection & Socketed Connection with as much default values as possible.

Tests were working in a way that the host (machine the tests are ran from) is connected to Dockerized MySQL server via TCP using Docker IP by default. 

So - in order to create tests for socketed connection I've come up with a socket -> socket tunnel via TCP. It works in a way that host creates socket, where the tests try to connect. This socket is only a traffic forwarder to 3307 port of the Docker instance of tested MySQL server. The docker instance has listener on port 3307 forwarding all traffic to mysqld socket. So there actually is a socketed connection - but relies on TCP. (Any ideas how to do socketed connection tests better while is the testing environment still being in Docker?)

The work was initiated by Issue #4

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -10,6 +10,7 @@ ADD test_database.sql /docker-entrypoint-initdb.d/
 RUN \
 apt-get update && apt-get install -y \
     patch \
+    socat \
     supervisor \
 && rm -rf /var/lib/apt/lists/*
 
@@ -21,6 +22,13 @@ RUN patch /usr/local/bin/docker-entrypoint.sh /docker-entrypoint.sh.patch
 # Since docker-entrypoint.sh drops privileges, we need to point supervisord
 # to a directory with required privileges. In this case, /tmp will suffice.
 ADD supervisord.conf /etc/supervisor/supervisord.conf
+
+# Enable mysqld socket (for testing purposes)
+RUN echo "socket = /var/run/mysqld/mysqld.sock" >> /etc/mysql/conf.d/mysql.cnf
+
+# Tunnel 3307 port to previously defined socket
+ADD mysql-socket-tunnel-3307 /usr/sbin/mysql-socket-tunnel-3307
+EXPOSE 3307
 
 # Ensure that we are using correct entrypoint due to wild development of
 # mysql:latest
--- a/db/mysql-socket-tunnel-3307
+++ b/db/mysql-socket-tunnel-3307
@@ -1,0 +1,2 @@
+#!/bin/sh
+exec socat TCP-LISTEN:3307,reuseaddr,fork UNIX-CLIENT:/var/run/mysqld/mysqld.sock
--- a/include/superior_mysqlpp/config.hpp
+++ b/include/superior_mysqlpp/config.hpp
@@ -1,0 +1,72 @@
+/*
+ * Author: Jiří Kozlovský
+ */
+
+#pragma once
+
+#include <string>
+
+#include <superior_mysqlpp/types/optional.hpp>
+
+namespace SuperiorMySqlpp
+{
+    struct SslConfiguration
+    {
+        const char* keyPath = nullptr;
+        const char* certificatePath = nullptr;
+        const char* certificationAuthorityPath = nullptr;
+        const char* trustedCertificateDirPath = nullptr;
+        const char* allowableCiphers = nullptr;
+    };
+
+    class ConnectionConfiguration {
+        private:
+            ConnectionConfiguration(const Optional<SslConfiguration>& _sslConfig, const std::string& _database, const std::string& _user, const std::string& _password,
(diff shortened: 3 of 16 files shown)


db/mysql-socket-tunnel-3307
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+socat TCP-LISTEN:3307,reuseaddr,fork UNIX-CLIENT:/var/run/mysqld/mysqld.sock
```
danopernis: Please use `exec socat ...` to save one process

.gitignore
```diff
@@ -12,6 +12,8 @@ debian/
 *.d
 *.o
 
+*.swp
```
danopernis: Please make this into separate commit

include/superior_mysqlpp/config.hpp
```diff
@@ -0,0 +1,75 @@
...
+                user{_user},
+                password{_password},
+                target{_target},
+                port{_port}
+            { }
+
+        public:
+
+            const bool hasSsl;
+            const struct SslConfiguration sslConfig;
```
danopernis: Please use `SuperiorMySqlpp::Optional<SslConfiguration>` instead of struct and bool

include/superior_mysqlpp/config.hpp
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            static ConnectionConfig getSocketConnectionConfig(const std::string& database, const std::string& user, const std::string& password,
+                    const std::string& socket="/var/run/mysqld/mysqld.sock")
+            {
+                    return ConnectionConfig{false, SslConfiguration{}, database, user, password, socket};
+            }
+
+            ConnectionConfig(const ConnectionConfig&) = default;
+            virtual ~ConnectionConfig() = default;
```
danopernis: I believe both members are redundant. Class is not intended for polymorphic usage and is indeed copy constructible

include/superior_mysqlpp/config.hpp
```diff
@@ -0,0 +1,75 @@
...
+    struct SslConfiguration
+    {
+        const char* keyPath = nullptr;
+        const char* certificatePath = nullptr;
+        const char* certificationAuthorityPath = nullptr;
+        const char* trustedCertificateDirPath = nullptr;
+        const char* allowableCiphers = nullptr;
+    };
+
+    class ConnectionConfig {
```
danopernis: Please rename this to ConnectionConfiguration for consistency with SslConfiguration

include/superior_mysqlpp/connection_def.hpp
```diff
@@ -103,6 +120,7 @@ namespace SuperiorMySqlpp
             return driver.getLoggerPtr();
         }
 
+
```
danopernis: Do not mix whitespace fixes with functional commits

tests/makefile
```diff
@@ -51,6 +51,10 @@ odr_object_files :=$(odr_source_files:.cpp=.o)
 odr_dependency_files :=$(odr_object_files:.o=.d)
 
 source_files :=$(sort $(shell find ./ -name "*.cpp" -not -name "*.odr.cpp"))
+
+# Uncomment this line to test connection only
+#source_files :=$(sort $(shell find ./ -name "*.cpp" -not -name "*.odr.cpp" | egrep "main|connection.cpp"))
```
danopernis: Why is this commented out?

tests/runtest.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if [ ! "$(dpkg -l | grep socat)" ]; then
```
danopernis: I would not bother with this, either it is installed and everything is OK, or it is not and script will fail anyway. You are not checking whether `docker` is installed, why would you special case socat?

danopernis (commented):
Nice work, please work on fixes, squash, rebase onto devel and I will review again.

db/Dockerfile
```diff
@@ -10,6 +10,7 @@ ADD test_database.sql /docker-entrypoint-initdb.d/
 RUN \
 apt-get update && apt-get install -y \
     patch \
+    socat \
```
jbouska567, line 13: what is this?

pnov (commented):
ok

majcl (changes requested):
Please add also description and example to Readme.md

smradCZ:
README.md is updated, please check it.

majcl (approved):

PR #27, opened 2017-04-05
Make Transaction dtor noexcept(false)

pnov:
Commit-query may throw an exception...


packages/fedora-22/libsuperiormysqlpp.spec
```diff
@@ -1,5 +1,5 @@
 Name:           libsuperiormysqlpp-dev
-Version:        0.2.2
+Version:        0.2.3
```
mody: It's an API change, I'd use version `0.3.0`

PR #30, opened 2017-06-29
Clang compile fixes

mody:
Most of the fixes are removing `const` but there is one fix to note: adding constexpr pow() function, see `to_integer.hpp`.


danopernis (approved):
Looks good! Can you please verify that gcc can still build this on all supported platforms?

mody:
All tests pass now. There was a mistake in 1st commit but it's fixed, rebased and pushed.

PR #31, opened 2017-07-18
fix compile-time pow()

mody:
Stupid bug went unnoticed because of missing tests for large interger types.


danopernis (approved):
Nasty bug

PR #34, opened 2017-10-11
Fix ostream operator for null values in Row

### LINKED ISSUE #32: Problem with `cout << row;`
Hi, i found problem, when i've try fetching data from database with simple query like this:

    while (Row row = result.fetchRow()) {
        std::cout << row << std::endl;
    }

This code fetch just part of first row. I assume that problem is in row.hpp on line 107, when are pushing results into stream. Problem may be solved calling instead `os << item;` use this `os << item.getStringView()`.

smradCZ:
It fixes #32. Reported issue was simulated on null values in table.
Some typos and some other details like space chars at end of lines were fixed in separate commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Please help us out by reporting bugs. (https://github.com/seznam/SuperiorMySqlpp
 We appreciate your feedback!
 
 # Preview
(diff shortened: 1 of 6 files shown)


smradCZ:
Closing. I'll move refactoring changes into separate branch.

PR #37, opened 2017-10-23
MariaDB has different internal representation of Timestamp

smradCZ:
MariaDB has Timestamp defined as UNSIGNED BINARY, but MySQL has it as BINARY without UNSINGED flag. This is the reason why `PreparedStatementTypeError` is thrown out from prepared statement validations.

I'd introduced 2 custom compiler parameters:
- TIMESTAMP_IGNORE_SIGN_VALIDATION - turn-off sign validation for Timestamp type
- TIMESTAMP_UNSIGNED_BINARY_TYPE - change Timestamp type to UNSIGNED

Moreover all tests are performed on MariaDB host too from now.


smradCZ:
Closing, signed/unsigned mark on Timestamp type is already ignored.

PR #39, opened 2017-11-14
Validation of uninitialized result binds

### LINKED ISSUE #14: Strange error while incorrect dynamic prepared statement result binding.
Example: 
1) We have two `unsigned int` columns in dynamic prepared statement result.
2) We incorrectly bind only first column, but twice, to different places.
3) Strange error occurs on updateResultBinding that `decimal` cannot be binded to `unsigned Long`
(the second type is ok - `unsigned int(10)` is actually `unsigned Long`, but this `decimal` is wrong, and also the error message misleading.

smradCZ:
It fixes #14 - Results which are not binded are checked as a first step of validation. Exception `PreparedStatementBindError` is thrown when some result is not initialized.

### DIFF
--- a/include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
+++ b/include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
@@ -381,6 +381,14 @@ namespace SuperiorMySqlpp
                     auto&& binding = *bindingsIt;
                     auto&& resultMetadata = *metadataIt;
 
+                    // It's needed to check buffer and length pointers to detect uninitialized binds.
+                    // Buffer may point to nullptr in case bind is initialized as ArrayBase of zero length
+                    // but the `length` will point to variable with buffer length in such case.
+                    if (bindingsIt->buffer == nullptr && bindingsIt->length == nullptr)
+                    {
+                        throw PreparedStatementBindError("Uninitialized bind for result at index " + std::to_string(index) + "!");
+                    }
+
                     auto&& bindingType = toFieldType(binding.buffer_type);
                     bool bindingIsUnsigned = binding.is_unsigned;
                     auto&& resultType = resultMetadata.getFieldType();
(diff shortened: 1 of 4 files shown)


include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -381,6 +381,14 @@ namespace SuperiorMySqlpp
                     auto&& binding = *bindingsIt;
                     auto&& resultMetadata = *metadataIt;
 
+                    // It's needed to check buffer and length pointers to detect uninitialized binds.
+                    // Buffer may point to nullptr in case bind is initialized as ArrayBase of zero length
+                    // but the `length` will point to variable with buffer length in such case.
+                    if (bindingsIt->buffer == nullptr && bindingsIt->length == nullptr)
```
majcl, line 387: Actually, we may need to ignore one or more result column - so the check might be probably for binding type, but then, this would need to have special ignored result type and also we must check, if underlaying mysql c-lib support ignoring result. This would require more logic, so i will create new issue for this.
Anyway shouldn't be enough to check only buffer length only?

majcl: Also, it's kind of strange to check binding engagement by this kind of hack :)

majcl (changes requested):

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
smradCZ: It's needed to check both conditions. The `length` can be `nullptr` in case the numeric type is binded.

See https://dev.mysql.com/doc/refman/5.7/en/c-api-prepared-statement-data-structures.html
> `length` is ignored for numeric and temporal data types because the `buffer_type` value determines the length of the data value.

majcl (approved):

Issue #38, opened 2017-11-03
Incorrect size passed to memset while clearing params/resultsBinding vectors


smradCZ:
In `dynamic_prepared_statement.hpp` clearing memory of param/result binding vector is not called properly because `sizeof(vector)` doesn't return size of allocated data but only size of vector instance (24 bytes ~ 3 pointers).

Moreover it seems that all these memset callings could be removed at all because the new values are zero-initialized when vector of PODs (plain old data) is resized (please correct me if I'm wrong).

danopernis:
You are indeed correct. Third argument to `std::memset` should be total number of bytes.

As for the second paragraph, standard requires `std:resize()` to *default-insert* the values. This is standardese for calling `allocator_traits<A>::construct(m, p)`. Allocator in question is `std::allocator`, therefore `::new((void *)p) A()` is called. Since no arguments to `new` are provided, after the allocation of memory the value is *default-initialized*. 

If I am not mistaken, after the POD is *default-initializated*, its value is not specified.

smradCZ:
Thank you for your reply.
As you said `::new((void *)p) A()` is called, it means the struct is *value-initialized* because there are parentheses after the template argument *A*.

I made some tests (compiled by GCC 7.1 and Clang 3.8 with --std=c++14) proving my theory.
Default initialization will be performed in case the `::new((void *)p) A` was called (without empty parentheses). Values are not specified in this particular case and they should have any value.

But I'm still not 100% sure about my opinion about that.

I would vote for using *std::memset* in every other case where the performance is not critical, but the performance is one of the keynotes of this library IMHO.

danopernis:
I stand corrected, `new` is called with initializer, albeit empty. Thus `value-initialization` is indeed performed.

PR #42, opened 2017-11-21
Remove redundant zeroing memsets

OpatrilPeter:
Fixes issue #38.
Removes redundant zeroing memsets - for std::vector. Also, memsets for std::array made redundant by the use of initializer.


majcl (changes requested):
LGTM,
 please just resolve conflicts.

majcl (approved):

PR #43, opened 2017-11-22
Fix incorrect Clang detection in test makefile 

OpatrilPeter:
Despite using C++ compiler for everything, detection of Clang was incorrectly using CC instead of CXX variable.


majcl (approved):
OK

PR #44, opened 2017-11-23
Remove -Wabi-tag from CXXFLAGS

smradCZ:
Mentioned GCC warning informs about situations where library may be
theoretically successfully linked with another one built with
incompatible ABI. For more info on this topic, see
https://developers.redhat.com/blog/2015/02/05/gcc5-and-the-c11-abi/)
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
Unfortunately, the only way to make this warning disappear is explicitly
put explicit abi_tag attribute to every single affected symbol, which
seems excessive and does not help readibility.
Notably, our internet research seem to suggest that this issue (linking
code built with old and new ABI) is almost never occuring in practice,
as all packages for given OS are by convention built with the same
version.
Judging from the relative lack of related problems, tutorials or general
discussion about this topic, -Wabi-tag seem to be now generally unused
by everyone.

This fix is one step from many in process of porting package to
debian-stretch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,3 +298,16 @@ DefaultLogger::setLoggerPtr(std::move(logger));
 
 # Current issues
 There are problems caused by MySQL C API's bad design which are solved by https://github.com/seznam/SuperiorMySqlpp/blob/master/include/superior_mysqlpp/low_level/mysql_hacks.hpp. This is causing problems with MariaDB which stripped down some symbols from their shared object that we use to fix this bug. (https://github.com/seznam/SuperiorMySqlpp/issues/2)
+
+## ABI tag warnings
+GCC supports `-Wabi-tag` that should warn when a type with ABI tag is used in context that not have that ABI tag. This warning should be used only for building shared libraries.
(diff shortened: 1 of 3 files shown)


majcl (changes requested):
Please, mention this change briefly in changelog, and also describe this (not so briefly :) ) in readme.

README.md
```diff
@@ -298,3 +298,16 @@ DefaultLogger::setLoggerPtr(std::move(logger));
 
 # Current issues
 There are problems caused by MySQL C API's bad design which are solved by https://github.com/seznam/SuperiorMySqlpp/blob/master/include/superior_mysqlpp/low_level/mysql_hacks.hpp. This is causing problems with MariaDB which stripped down some symbols from their shared object that we use to fix this bug. (https://github.com/seznam/SuperiorMySqlpp/issues/2)
+
+## ABI tag warnings
+GCC supports `-Wabi-tag` that should warn when a type with ABI tag is used in context that not have that ABI tag. This warning should be used only for building shared libraries.
```
majcl: Please add reference to documentation.

majcl (changes requested):

majcl (approved):

PR #45, opened 2017-11-27
GCC 6.3 warning - throwing an exception from noexcept method

### LINKED ISSUE #41: Incorrect functions called to get error info when mysql_stmt_close fails
When `mysql_stmt_close()` fails the information about error should be obtained from `mysql_error()`/`mysql_errno()` calling instead of `mysql_stmt_error()`/`mysql_stmt_errno()`.

The only occurrence in code base is located in *include/superior_mysqlpp/low_level/dbdriver.hpp* in method *Statement::close()*.

See https://dev.mysql.com/doc/refman/5.7/en/mysql-stmt-close.html for more info.

smradCZ:
Error will be logged instead of throwing an exception from noexcept
method Statement::close() and program will be terminated by
calling std::terminate directly.
It's fixed because is not intended to throw messages directly from
methods with noexcept specifier and this is also checked by compilers
with -Wall option by default.

It fixes #41 also.

This is one step from many in process of porting package to debian-stretch.

### DIFF
--- a/include/superior_mysqlpp/logging.hpp
+++ b/include/superior_mysqlpp/logging.hpp
@@ -56,7 +56,7 @@ namespace SuperiorMySqlpp {
          */
 
         /*
-         * You are to choose to derive either from class Interface or class Empty
+         * You are to choose to derive either from class Interface or class Base
          * so you shall either implement all logging functions or only those you want.
          */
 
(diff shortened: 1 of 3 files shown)


include/superior_mysqlpp/logging.hpp
```diff
@@ -703,7 +711,7 @@ namespace SuperiorMySqlpp {
         }
 
     public:
-        static bool isDestoyed()
+        static bool isDestroyed()
```
majcl: This is kind of unrelated change. Please put it into another pull request. 
Also, this is unfortunately interface change and thus this should implicate library major version update. Though, for now, we still have major version 0 - so this probably is not necessary. 
Anyway, my first request stands. Please put this into another pull request.

packages/_debian-common/changelog
```diff
@@ -2,6 +2,7 @@ libsuperiormysqlpp (0.3.1) UNRELEASED; urgency=medium
 
   [ Radek Smejdir ]
   * Fix ostream operator for null values in Row
+  * noexcept Statement.close() will not throw MysqlInternalError anymore
```
majcl: Please mention also, that instead log function will be called and std::terminate will be called.

majcl (changes requested):

majcl (approved):

PR #47, opened 2017-11-29
 Replace `string_view::to_string` with string ctor 

OpatrilPeter:
Method `to_string` in experimental (pre-C++17) version of `std::string_view` was removed from C++17 standard. New preferred (and backward compatible) way is calling std::string constructor.

### DIFF
--- a/include/superior_mysqlpp/query_result.hpp
+++ b/include/superior_mysqlpp/query_result.hpp
@@ -115,7 +115,7 @@ namespace SuperiorMySqlpp
                 }
                 else
                 {
-                    throw OutOfRange{"Column name \"" + columnName.to_string() + "\" not found!"};
+                    throw OutOfRange{"Column name \"" + std::string(columnName) + "\" not found!"};
                 }
             }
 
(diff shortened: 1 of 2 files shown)


include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -56,7 +56,7 @@ namespace SuperiorMySqlpp
                 while (destination != end)
                 {
                     endpoint = *destination++;
-                    addresses.emplace_back(endpoint.address().to_string());
+                    addresses.emplace_back(std::string(endpoint.address()));
```
majcl: This is boost::asio::ip::address class and it has to_string() member method. Are you really sure, you don't want to use it? Why? It has nothing to do with std:string_view.

majcl (changes requested):

include/superior_mysqlpp/dnsa_connection_pool.hpp
OpatrilPeter: You were indeed correct, that was an oversight and shall be already fixed

majcl (approved):

PR #49, opened 2017-12-11
Fix method typo isDestoyed -> is Destroyed

smradCZ:
This change breaks `DefaultLogger` interface and that should be reflected by increasing major version of package (or maybe minor version at least in case we haven't released version 1.0.0 yet).

Please update version number accordingly to this change when next release of package will be released, thanks.

### DIFF
--- a/packages/_debian-common/changelog
+++ b/packages/_debian-common/changelog
@@ -7,6 +7,10 @@ libsuperiormysqlpp (0.4.0) UNRELEASED; urgency=medium
   * Extend psReadValues exceptions
   * feat: Detect lost connection when using query
 
+  [ Radek Smejdir ]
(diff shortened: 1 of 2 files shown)


packages/_debian-common/changelog
```diff
@@ -7,6 +7,9 @@ libsuperiormysqlpp (0.4.0) UNRELEASED; urgency=medium
   * Extend psReadValues exceptions
   * feat: Detect lost connection when using query
 
+  [ Radek Smejdir ]
+  * Fix typo DefaultLogger::isDestoyed -> DefaultLogger::isDestroyed
```
OpatrilPeter: I'd mention that original method is still there, deprecated.

OpatrilPeter (approved):

PR #51, opened 2017-12-14
Packaging support for Debian Stretch

smradCZ:
This PR adds support for building Debian Stretch package.

Clang is used for compilation because AddressSanitizer included in G++6.3 reports false-positive 1040B leak. False-positive leak because when it's compiled by G++8.0 with newer version of AddressSanitizer or by Clang it doesn't report any memory leak as well as valgrind's memcheck.

The `mysql_hacks.hpp` are no longer needed from MySQL v5.7.9 (5.7.20 available on Oracle's official repo for debian). MariaDB stripped symbols needed for mysql_hacks and still have memory leaks when we move connection between threads. MariaDB and related stuff will be part of another PR.


mody:
By saying:
> The mysql_hacks.hpp are no longer needed from MySQL v5.7.9

you mean they fixed it at Oracle?

smradCZ:
Yes, see this commit https://github.com/mysql/mysql-server/commit/13ccce6f380844fd030e33a06be10afaa91d56c2
Basically they hide these symbols in release version of lib

majcl (approved):

PR #59, opened 2018-01-09
Add Stmt tieResults convenience method

OpatrilPeter:


smradCZ (commented):
Please add example of `tieResult` method to README.md and resolve conflicts in debian/changelog

OpatrilPeter:
After giving it some thought and team discussion, I've decided to drop this PR.
While it is more concise then usual std::tie/std::get solution, I feel it's less suprising to just use the usual ways of dealing with std::tuple, especially in face of C++17 structured bindings.

PR #62, opened 2018-02-12
Add commentary for prepared statement and related code

OpatrilPeter:
Documentation for PreparedStatement machinery.
Things still requiring review is annotated by !!!!!.
Correct generation of documentation by Doxygen is not properly tested yet.

### DIFF
--- a/include/superior_mysqlpp/prepared_statement.hpp
+++ b/include/superior_mysqlpp/prepared_statement.hpp
@@ -24,10 +24,89 @@
 #include <superior_mysqlpp/traits.hpp>
 #include <superior_mysqlpp/types/tags.hpp>
 
+/************************************
+ * Common usages of PreparedStatement
+ ************************************
+ *
+ *
+ *  Notes - We prefer 'Sql::Int' over 'int' for portability.
+ *          Empty ResultBindings can be ommited entirely.
+ *          Storage of parameter values is always inside prepared statement, even if lvalues are passed for initialization.
+ *          Parameter values are passed in varargs fashion and type of ParamBindings is inferred from that.
+ *            So, you are generally expected to provide suitable values of parameters at a time of construction.
+ *          Additional template parameters of makePreparedStatement allow further customization.
+ *
+ *
+ *  Step - creating prepared statement
+ *
+ * ``` c++
+ * SuperiorMySqlpp::Connection connection;
+ * ...
+ *
+ * // case a) No parameters
+ * auto preparedStatement = connection.makePreparedStatement<ResultBindings<Sql::Int, Sql::Int>>(
+ *     "SELECT `id`, `money` FROM `transactions`"
+ * );
+ * // case b) No result, 4 parameters
+ * auto preparedStatement = connection.makePreparedStatement(
+ *     "INSERT INTO `tablename` VALUES (?, ?), (?, ?)",
+ * 	42, Nullable<StringDataBase<12>>{"Hello"},
+ * 	43, Nullable<StringDataBase<12>>{"world"}
+ * );
+ * // case c) Deferred parameters - using PreparedStatement ctor directly
+ * auto preparedStatement = PreparedStatement<ResultBindings<Sql::TinyInt>,ParamBindings<Sql::Int, Sql::Int>>(
+ * 	connection,
+ *     "INSERT INTO `tablename` VALUES (?, ?)"
+ * );
+ * ```
+ *
+ *  Step - execution
+ * ``` c++
+ * // This actually performs the query on the server.
+ * // May be called multiple times
+ *
+ * preparedStatement.execute();
+ * ```
+ *
+ *  Step - data retrieval
+ *    - needed only for SELECT(-like) statements
+ *
+ * ``` c++
+ * // Fetches single row into prepared statement's storage
+ * while (preparedStatement.fetch()) // This is true as long as data are available (throws on errors like truncation)
+ * {
+ * 	Sql::Int id, money;
+ *
+ * 	// Now, one can read std::tuple within prepared statement directly
+ * 	std::tie(id, money) = preparedStatement.getResult();
+ * }
+ * ```
+ *
+ * // Optional step - if we decide to change parameters (shall happen before execution)
+ * ``` c++
+ * preparedStatement.setParams(123,StringData{"Xyzzy"});
+ * preparedStatement.updateParamsBindings(); // Updates binding information, required step for most nontrivial parameter types
+ * preparedStatement.execute(); // Statement is performed now
+ * preparedStatement.setParams(666,StringData{"Elbereth"});
+ * preparedStatement.updateParamsBindings();
+ * preparedStatement.execute(); // Statement is performed again, now with different parameters
+ * ```
+ **/
 
 
 namespace SuperiorMySqlpp
 {
+    /**
+     * @brief Bindings of storage for either parameters (inputs) or results (outputs) of SQL queries.
+     * Bindings class encapsulates the storage of heterogenous data needed for SQL parameters and results.
+     * The underlying storage itself is in std::tuple member data, auxiliary binding information is stored
+     * in array of MYSQL_BIND structures used from underlying C mysql client -- member bindings.
+     * @tparam IsParamBingngs - Special flag for template specialization.
+     *                          If Bindings is used for query parameter, it is true.
+     *                          If Bindings is used for query result, it is false.
+     * @tparam Types - Parameter pack of types forming storage of individual fields.
+     * @remark In user code, prefer use convenience aliases ParamBindings and ResultBindings.
+     */
     template<bool IsParamBinding, typename... Types>
     class Bindings
     {
@@ -35,13 +114,19 @@ namespace SuperiorMySqlpp
         static constexpr auto kArgumentsCount = sizeof...(Types);
         static constexpr bool isParamBinding = IsParamBinding;
 
+        // Underlying storage for binding's data.
         std::tuple<Types...> data;
+        // Array of binding structures pointing to storage in member #data.
         std::array<MYSQL_BIND, sizeof...(Types)> bindings;
 
 
     public:
-        /* Note: using just 'bindings{}' triggers (premature) warning in GCC 4.9.2,
-         * see discussion at https://stackoverflow.com/questions/31271975/why-can-i-initialize-a-regular-array-from-but-not-a-stdarray */
+        /*
+         * Bindings constructor.
+         * @params args Args is a pack of values used for initialization of binding's storage.
+         * @remark using just 'bindings{}' triggers (premature) warning in GCC 4.9.2,
+         *         see discussion at https://stackoverflow.com/questions/31271975/why-can-i-initialize-a-regular-array-from-but-not-a-stdarray
+         */
         template<typename... Args>
         Bindings(Args&&... args)
             : data{std::forward<Args>(args)...}, bindings{{}}
@@ -56,19 +141,51 @@ namespace SuperiorMySqlpp
         Bindings& operator=(Bindings&&) = default;
         ~Bindings() = default;
 
+        /*
+         * Updates bindings to properly reflect on data.
+         * Is fundamental for useability of Bindings and is thus called during initialization.
+         */
         void update()
         {
             detail::initializeBindings<IsParamBinding, Types...>(bindings, data);
         }
     };
 
+    /**
+     * ParamBindings specifies bindings that shall be used for query parameters.
+     * @tparam Types List of types matching query params.
+     */
     template<typename... Types>
     using ParamBindings = Bindings<true, Types...>;
 
+    /**
+     * ResultBindings specifies bindings that shall be used for query results.
+     * @tparam Types List of types matching query results.
+     */
     template<typename... Types>
     using ResultBindings = Bindings<false, Types...>;
 
-
+    /**
+     * @brief Static MySQL prepared statement.
+     * Static prepared statements need to specify parameter and result bindings during creation,
+     * values of them are stored internally.
+     * Please note that just because your query has parameters, you do not neccessarily need
+     * parameter bindings, you can often hardcode or concatenate them into query string.
+     * Two advantages compared to using SuperiorMySqlpp::Query are binary communication with server
+     * and stronger type safety guarantees.
+     * Created by constructor or Connection::makePreparedStatement.
+     * @tparam ResultBindings Specialization of #ResultBindings.
+     * @tparam ParamBindings Specialization of #ParamBindings.
+     * @tparam storeResult If true, fetches results in "store mode", whole result set in one operation.
+     *                     If false, fetches results in "use mode", incrementally.
+     *                     Both modes offers some additional unique methods.
+     * @tparam validateMode Defines conversion class, i.e. how strictly must match type
+     *                      provided in result/param storage and the one from table metadata.
+     *                      If stronger class of conversion is required than the one performed, it leads to error.
+     * @tparam warnMode Defines level of conversion that triggers warning, but is still acceptable.
+     * @tparam ignoreNullable If true, one can store nullable data of type T from database to result storage of type T (directly withnout SuperiorMySqlpp::Nullable).
+     *                      If the nullable is in null state, nothing will be assigned to given result field.
+     */
     template<typename ResultBindings, typename ParamBindings, bool storeResult, ValidateMetadataMode validateMode, ValidateMetadataMode warnMode, bool ignoreNullable>
     class PreparedStatement : public detail::PreparedStatementBase<storeResult, validateMode, warnMode, ignoreNullable>
     {
@@ -82,6 +199,16 @@ namespace SuperiorMySqlpp
         bool hasValidatedResultMetadata = false;
 
     private:
+        /**
+         * @brief Constructor for PreparedStatement.
+         * This overload is designated by SuperiorMySqlpp::fullInitTag and specifies both parameters and results as tuples.
+         * @remark It is used internally as implementation of matching public constructor.
+         * @param driver LowLevel::DBDriver wrapping MySql connection.
+         * @param query String containing query to be executed.
+         * @param tag Tag field for specifying this overload.
+         * @param resultArgsTuple Forwarded std::tuple containing initial values of resultBindings.
+         * @param paramArgsTuple Forwarded std::tuple containing initial values of paramBindings.
+         */
         template<typename ResultArgsTuple, typename ParamArgsTuple, std::size_t... RI, std::size_t... PI>
         PreparedStatement(LowLevel::DBDriver& driver,
                           const std::string& query,
@@ -110,6 +237,15 @@ namespace SuperiorMySqlpp
         }
 
     public:
+        /**
+         * @brief Constructor for PreparedStatement.
+         * This overload is designated by SuperiorMySqlpp::fullInitTag and specifies both parameters and results as tuples.
+         * @param driver DBDriver wrapping MySql connection.
+         * @param query String containing query to be executed.
+         * @param tag Tag field for specifying this overload.
+         * @param resultArgsTuple Forwarded std::tuple containing initial values of resultBindings.
+         * @param paramArgsTuple Forwarded std::tuple containing initial values of paramBindings.
+         */
         template<typename... RArgs, typename... PArgs, template<typename...> class ResultArgsTuple, template<typename...> class ParamArgsTuple>
         PreparedStatement(LowLevel::DBDriver& driver,
                           const std::string& query,
@@ -127,6 +263,13 @@ namespace SuperiorMySqlpp
         {
         }
 
+        /**
+         * @brief Constructor for PreparedStatement.
+         * This overload is for useful when only DBDriver and (parametrized) query are needed.
+         * @param driver DBDriver wrapping MySql connection.
+         * @param query String containing query to be executed.
+         * @param params Forwarded varargs used for initialization of paramBindings.
+         */
         template<typename... Args>
         PreparedStatement(LowLevel::DBDriver& driver, const std::string& query, Args&&... params)
             : PreparedStatement
@@ -139,6 +282,16 @@ namespace SuperiorMySqlpp
             InitializeResult(resultBindings.data);
         }
 
+        /**
+         * @brief Constructor for PreparedStatement.
+         * This overload is useful when #Connection is available.
+         * Uses fullInitTag and specifies both parameters and results as tuples.
+         * @param connection Reference to opened MySql connection.
+         * @param query String containing query to be executed.
+         * @param tag Tag field for specifying this overload.
+         * @param resultArgsTuple Forwarded std::tuple containing initial values of resultBindings.
+         * @param paramArgsTuple Forwarded std::tuple containing initial values of paramBindings.
+         */
         template<typename ResultArgsTuple, typename ParamArgsTuple>
         PreparedStatement(Connection& connection,
                           const std::string& query,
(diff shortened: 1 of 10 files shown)


include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -28,6 +28,17 @@
 
 namespace SuperiorMySqlpp
 {
+    /**
+     * @brief Bindings of storage for either parameters (inputs) or results (outputs) of SQL queries.
+     * Bindings class encapsulates the storage of heterogenous data needed for SQL parameters and results.
+     * The binding info itself is stored in std::tuple @a data, auxiliary binding information is stored in array of
```
majcl: Actually bindings storage is stored in std::tuple data, isn't it?

majcl: Isn't it possible to ref data(in "...std::tuple @a data, auxili..." by #data or something similar, instead of manual highliting with \@a?

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -28,6 +28,17 @@
...
+     * MYSQL_BIND structures used from underlying C mysql client -- member @a bindings.
+     * @tparam IsParamBingngs - Special flag for template specialization.
+     *                          If Bindings is used for query parameter, it is true.
+     *                          If Bindings is used for query result, it is false.
+     * @tparam Types - Parameter pack of types forming storage of individual fields.
```
majcl: storage *for* individual... ?

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -40,8 +51,12 @@ namespace SuperiorMySqlpp
 
 
     public:
-        /* Note: using just 'bindings{}' triggers (premature) warning in GCC 4.9.2,
-         * see discussion at https://stackoverflow.com/questions/31271975/why-can-i-initialize-a-regular-array-from-but-not-a-stdarray */
+        /*
+         * Bindings constructor.
+         * @params args Args are used for tuple initialization.
```
majcl: Probably should be specified, that args are used as a storage initialization.

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -56,19 +71,47 @@ namespace SuperiorMySqlpp
...
+    /*
+     * @brief Static MySQL prepared statement.
+     * Static prepared statements needs to know the parameter bindings.
+     * Two advantages compared to using #Query - binary communication with server
+     * and stronger type safety guarantees.
+     * Created by constructor or Connection::makePreparedStatement.
+     * @tparam ResultBindings Specialization of #ResultBindings.
+     * @tparam ParamBindings Specialization of #ParamBindings.
+     * @tparam storeResult If true, fetches results in "store mode" -- whole result set in one operation.
+     *                     If false, fetches results in "use mode", incrementally.
```
majcl: This two lines should probably have similar formating ;-)

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -56,19 +71,47 @@ namespace SuperiorMySqlpp
...
+     *                     Both modes offers some additional unique methods.
+     * @tparam validateMode Defines validation strategy, i.e. how strict type control is.
+     *                      If stronger class of conversion is required, it leads to error.
```
majcl: conversion class

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -56,19 +71,47 @@ namespace SuperiorMySqlpp
...
+     * @tparam warnMode Defines level of conversion that triggers warning, but is accepted.
+     * @tparam ignoreNullable If true, one can store #Nullable<T> in storage of type T. If the nullable is empty,
```
majcl: I would rephrase - if field is actually NULL, the storage will remain unchanged...

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -139,6 +207,16 @@ namespace SuperiorMySqlpp
             InitializeResult(resultBindings.data);
         }
 
+        /**
+         * @brief Constructor for PreparedStatement.
+         * This overload is for useful when Connection is available.
```
majcl: Link to Connection

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -56,19 +71,47 @@ namespace SuperiorMySqlpp
...
+     * @tparam ResultBindings Specialization of #ResultBindings.
+     * @tparam ParamBindings Specialization of #ParamBindings.
+     * @tparam storeResult If true, fetches results in "store mode" -- whole result set in one operation.
+     *                     If false, fetches results in "use mode", incrementally.
+     *                     Both modes offers some additional unique methods.
+     * @tparam validateMode Defines validation strategy, i.e. how strict type control is.
+     *                      If stronger class of conversion is required, it leads to error.
+     * @tparam warnMode Defines level of conversion that triggers warning, but is accepted.
+     * @tparam ignoreNullable If true, one can store SuperiorMySqlpp::Nullable<T> in storage of type T. If the nullable is empty,
+     *                        nothing will be assigned.
```
majcl: What does mean if nullable is empty?

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -185,42 +274,72 @@ namespace SuperiorMySqlpp
...
+         * Accessor of tuple containing storage of result bindings.
+         * Const variant.
+         */
         const auto& getResult() const noexcept
         {
             return resultBindings.data;
         }
 
+        /**
+         * Convenience function to replace current query params.
```
majcl: What does it exactly mean replace current query params? And what about connection to updateParamsindings? Please explain this more thoroughly.

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -185,42 +274,72 @@ namespace SuperiorMySqlpp
...
+         * @param args Forwarded new data,
+         */
         template<typename... Args>
         void setParams(Args&&... args)
         {
             paramsBindings.data = std::forward_as_tuple(std::forward<Args>(args)...);
         }
 
+        /**
+         * Returns size of parameterBindings.
```
majcl: Is it really size or count of binded parameters?

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -185,42 +274,72 @@ namespace SuperiorMySqlpp
...
+         */
         static constexpr auto getParamsCount() noexcept
         {
             return decltype(paramsBindings)::kArgumentsCount;
         }
 
+        /**
+         * Returns size of resultBindings.
```
majcl: Is it really size or count of binded results?

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -185,42 +274,72 @@ namespace SuperiorMySqlpp
...
+         */
         static constexpr auto getResultCount() noexcept
         {
             return decltype(resultBindings)::kArgumentsCount;
         }
 
+        /**
+         * @brief Executes prepared statement.
+         * This method actually performs given query, after validating its metadata.
```
majcl: Isn't it the other way around? First performs query and then validates actual result metadata with result bindings?

include/superior_mysqlpp/prepared_statements/binding_types.hpp
```diff
@@ -25,10 +25,19 @@ namespace SuperiorMySqlpp
             Timestamp,
         };
 
-
+        /**
+         * This boolean trait describes if type T can be uaed as a storage for
```
majcl: typo: uaed => used

include/superior_mysqlpp/prepared_statements/binding_types.hpp
```diff
@@ -25,10 +25,19 @@ namespace SuperiorMySqlpp
...
+         * a given binding type (of enum type #BindingTypes).
+         * This is a variant for query parameters and defaults to false.
+         */
         template<BindingTypes, typename T>
         struct CanBindAsParam : std::false_type {};
 
+        /**
+         * This boolean trait describes if type T can be uaed as a storage for
```
majcl: typo uaed => used

include/superior_mysqlpp/prepared_statements/default_initialize_result.hpp
```diff
@@ -10,8 +10,10 @@
 
 namespace SuperiorMySqlpp
 {
-    /*
-     * initializeResultItem
+    /**
+     * Generic template and default implementation of initialization for one field of result binding.
```
majcl: This needs to be explained more throroughly. Where and how and why this is used.

include/superior_mysqlpp/prepared_statements/get_binding_type.hpp
```diff
@@ -16,27 +16,44 @@ namespace SuperiorMySqlpp
 {
     namespace detail
     {
+        /**
+         * Mapping of native C++ types to fitting field type
+         * @return FieldType enumeration of matching type
+         * !!!!! This code is incorrect - sql types are exact size, C++ are architecture dependent
```
majcl: Correct note, but, is underlying C mysql client even available for different architectures?

include/superior_mysqlpp/prepared_statements/get_binding_type.hpp
```diff
@@ -16,27 +16,44 @@ namespace SuperiorMySqlpp
 {
     namespace detail
     {
+        /**
+         * Mapping of native C++ types to fitting field type
```
majcl: I would add to *mysql* field type. Also somewhere should be added, that it is preffered to use *mysql* types instead of native types, wherever possible.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -14,14 +14,23 @@
...
+         * Trait specialization saying that std::string can be used as parameter
+         * for string field type.
+         */
         template<>
         struct CanBindAsParam<BindingTypes::String, std::string> : std::true_type {};
 
 
-        /*
-         * Common bindings
+        /**
```
majcl, line 42: It would be helpful here, to have more thorough description, what is done in the function and that this is only internal part of actual initialization(because Nullable<T> must have also initialize its "content" T)

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -45,6 +54,9 @@ namespace SuperiorMySqlpp
             binding.is_unsigned = std::is_unsigned<PureType_t>()? true : false;
         }
 
+        /**
+         * Param binding initialization specialization for floats.
```
majcl, line 72: Please mention in doc also long double exception and possibly the reason.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -14,14 +14,23 @@
 
 namespace SuperiorMySqlpp
 {
+    /*
+     * Objective of this whole file is proper initialization of bindings.
```
majcl, line 18: Also please describe a little how it is actually designed. It goes a long way, when someone needs to add something. Also when and how are these initializations called.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -173,48 +229,68 @@ namespace SuperiorMySqlpp
...
-         * Initialize binding implementation
+        /**
+         * Initializes individual binding.
+         * Generic template, actual implementation is in specializations.
          */
         template<bool isParamBinding>
         struct InitializeBindingImpl;
 
+        /**
+         * Specialization of #InitializeBindingImpl for query parameters.
```
majcl, line 309: Explain please the specialization by true/false template argument

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -270,8 +355,10 @@ namespace SuperiorMySqlpp
         };
 
 
-        /*
-         * Initialize Bindings
+        /**
```
majcl, line 375: This is probably the only "importantly-visible exported" method from this module. Please describe also in java doc, how does it works.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -68,23 +94,39 @@ namespace SuperiorMySqlpp
             }
 
         protected:
+            /**
+             * This method implements the store() specialized behavior.
```
majcl: What does actually mean "specialized behaviour"?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -68,23 +94,39 @@ namespace SuperiorMySqlpp
...
+             */
             void storeOrUse()
             {
                 this->statement.storeResult();
             }
         };
 
+        /**
+         * Specialization of StatementBase where statement results are "used" - not
+         * fetched all at one.
```
majcl: The code design, how prepared statement distinguishes between *store* and *use* results should be explained somewhere.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -68,23 +94,39 @@ namespace SuperiorMySqlpp
...
+             * As this is the default mode for mysql client, nothing needs to be done.
+             */
             void storeOrUse() const
             {
             }
 
+            /**
+             * Returns the amount of fetched rows.
+             * @return Amount of already fetched rows.
+             * !!!!! Underlying mysql_stmt_num_rows does NOT gurantee such behavior!
```
majcl: This returns already fetched rows, not total result rows, or even this lighter guarantee isn't right?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -155,7 +203,11 @@ namespace SuperiorMySqlpp
             }
 
         public:
-
+            /**
+             * Returns metadata for the statement result.
+             * There metadata are retrieveable even before execute().
```
majcl, line 219: Are you sure they are available before execute()?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -192,45 +254,71 @@ namespace SuperiorMySqlpp
                 return ok;
             }
 
-            /**
-             * TODO: Statement.fetchColumn is void and can throw
```
majcl, line 196: Why this comment is deleted?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -192,45 +254,71 @@ namespace SuperiorMySqlpp
...
+             * @return void
+             * !!!!! Should return void, this form is misleading.
+             */
             auto sendLongData(unsigned int paramNumber, const std::string& data)
             {
                 return this->statement.sendLongData(paramNumber, data);
             }
 
+            /**
+             * Sets how many rows shall be prefetched in single fetch call.
```
majcl, line 305: Does it have meaning for both *use* and *store* results?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -192,45 +254,71 @@ namespace SuperiorMySqlpp
...
+             */
             unsigned long getPrefetchRowCount()
             {
                 unsigned long result;
                 this->statement.getAttribute(LowLevel::DBDriver::Statement::Attributes::prefetchRows, result);
                 return result;
             }
 
+            /**
+             * If related attribute is set to true, the metadata for field's metadata is updated on store.
```
majcl: What metadata and how and when they are updated?

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
```diff
@@ -240,6 +328,10 @@ namespace SuperiorMySqlpp
 
 
         public:
+            /**
+             * Validates metadata for given ResultBindings compared to metadata recieved from the server.
```
majcl, line 352: Just mention please, when it is called.

majcl (changes requested):
I like the documentation adding. I hope it will be really helpful in future development. But there is one thing i miss here: Somewhat thorough description of how ParamsBindings and ResultBindings works in PreparedStatements and DynamicPreparedsStatement. 
When the storage of provided arguments is used and when the provided argument's values are only copied.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: This comment is deleted because this function was in fact not working at all - and is rewritten along with proper documentation in following pull request #63.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: According to MySql documentation, yes. It calls `mysql_stmt_result_metadata` that fetches this information independently on `execute()`, see example at https://dev.mysql.com/doc/refman/5.7/en/mysql-stmt-fetch.html.

include/superior_mysqlpp/prepared_statement.hpp
OpatrilPeter: By empty I meant null -- I like to consider Nullable equivalent of Optionals. Do you have any better way to describe it in mind?

include/superior_mysqlpp/prepared_statement.hpp
OpatrilPeter: It is count indeed. IMO, because in C++ it is common that size() returns the amount of elements for std containers, I consider size and count to be equivalent terms. Are you considering ambiguity with `sizeof`-like semantics?

include/superior_mysqlpp/prepared_statement.hpp
OpatrilPeter: See preceding discussion.

include/superior_mysqlpp/prepared_statement.hpp
OpatrilPeter: Good catch! The actual order of operations here is following:
`mysql_stmt_execute()` <- this preforms the query, producing the result set (for SELECT-like statements)
*validation*
(optional step for `store()`) `mysql_stmt_store_result()` <- this, however, performs the "heavy lifting" of delivery

include/superior_mysqlpp/prepared_statements/binding_types.hpp
OpatrilPeter: fixaed ;)

include/superior_mysqlpp/prepared_statements/get_binding_type.hpp
OpatrilPeter: I haven't seen any specific support in C client for more exotic architectures, but I feel that being more explicit in data type choice (with no extra cost) is a defensive step in right direction nontheless -- I suggest leaving this one kept as it is - in following pull request, this code was rewritten, so we can debate over that specific change more concretely and productively.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: Will clarify in more detailed description. FYI, the same method exist for subclass that has *use*() strategy and for subclass that does *store*() strategy - it does nothing for *use*, but calls mysql_stmt_store_result for *store*.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: It is not guaranteed explicitly in documentation. Further research needs to be made into source code if such behaviour is explicitly mentioned at least there or if it is purely implemntation detail.

include/superior_mysqlpp/prepared_statement.hpp
OpatrilPeter: I clarified the meaning a little - mainly the fact that this flag is about storing results.

include/superior_mysqlpp/prepared_statements/default_initialize_result.hpp
OpatrilPeter: As far as I can see, this function itself is only called from function right below - most likely it should be inside detail:: namespace. That said, it actually calls specializable function InitializeResultItemImpl, which does nothing in general variant, but can be specialized for arbitrary type - this information is already provided in comment above.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: After looking harder, this one seems problematic. According to https://dev.mysql.com/doc/refman/5.7/en/mysql-stmt-attr-set.html, it does following: Number of rows to fetch from server at a time _when using a cursor_.  MariaDB does not mention that at all. We don't even have required method (outside DBDriver) to create cursors.

For now, I'll only add commentary about this.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: I did some testing, as basic tests did not cover this.
After calling this method in use() mode, it always returns 0 for me (with debian-jessie default mysql client). Was that ever tested at all?

I'll add suggestion for removal.

OpatrilPeter: Alternatively, correct implementation can be created by simple counter addition in that specialization.

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
OpatrilPeter: I hope description of tparam for all related classes and in general usage description will be enough.

include/superior_mysqlpp/prepared_statements/validate_metadata_modes.hpp
```diff
@@ -44,6 +50,12 @@ namespace SuperiorMySqlpp
         }
     }
 
+    /**
+     * Function returns for given field type "typeId" shared by same types
+     * @param type
+     * @return Arbitary integer that is however unique for each group of same types
+     * !!!!!
```
OpatrilPeter: Request for feedback, this part is a guesswork.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: I added this information to the general version.

include/superior_mysqlpp/prepared_statement.hpp
```diff
@@ -56,19 +141,48 @@ namespace SuperiorMySqlpp
...
+     * ResultBindings specifies bindings that shall be used for query results.
+     * @tparam Types List of types matching query results.
+     */
     template<typename... Types>
     using ResultBindings = Bindings<false, Types...>;
 
-
+    /**
+     * @brief Static MySQL prepared statement.
+     * Static prepared statements needs to know the parameter bindings.
```
majcl: Actually there is no need to have parameters, the query can be just one string as in normal query, but possible results must be bounded to some storage..

include/superior_mysqlpp/prepared_statements/prepared_statement_base.hpp
majcl: Yes, this should probably be then counted manually in use spec.

majcl (approved):

PR #68, opened 2018-03-02
Fix exact int types

OpatrilPeter:
As SQL uses exact-width integer types, using native C ints may cause portability issues - this PR rewrites such occurences.

Note that this currently causes warnings in test suite.
They are not errors (only hint at usage of inappropriate type) but probably should be fixed.

EDIT:
After review, this PR scope changed - native C ints are not arbitrary, but are explicitly specified in underlying C library, so this PR will be limited to highlighting that fact.

### DIFF
--- a/include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
+++ b/include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
@@ -65,7 +65,7 @@ namespace SuperiorMySqlpp
 
             binding.buffer = &value;
             binding.buffer_type = detail::toMysqlEnum(getBindingType<Signed_t>());
-            binding.is_unsigned = std::is_unsigned<PureType_t>()? true : false;
+            binding.is_unsigned = std::is_unsigned<PureType_t>::value;
         }
 
         /**
--- a/include/superior_mysqlpp/prepared_statements/binding_types.hpp
+++ b/include/superior_mysqlpp/prepared_statements/binding_types.hpp
@@ -13,6 +13,11 @@ namespace SuperiorMySqlpp
 {
     namespace detail
     {
+        /**
+         * Enumeration of various kinds of bindings.
+         * Integral and floating types are ommited, those types
+         * are referred to by appropriate primitive C type instead.
+         */
         enum class BindingTypes
         {
             Nullable,
--- a/include/superior_mysqlpp/prepared_statements/get_binding_type.hpp
+++ b/include/superior_mysqlpp/prepared_statements/get_binding_type.hpp
@@ -19,13 +19,14 @@ namespace SuperiorMySqlpp
         /**
          * Mapping of native C++ types to fitting MySQL field type.
          * @return FieldType enumeration of matching type.
-         * @remark This code is not completely portable - SQL types have exact size, C++ ones are architecture dependent.
+         * @remark Mapping between MySql and C types is based on specifications from underlying library
+         *         at https://dev.mysql.com/doc/refman/5.7/en/c-api-prepared-statement-type-codes.html
          */
         template<typename T>
         inline constexpr FieldTypes getBindingType() = delete;
 
         /**
-         * @remark Specialization for signed char, see generic version.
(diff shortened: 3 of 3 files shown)


include/superior_mysqlpp/sql_types.hpp
```diff
@@ -15,17 +15,18 @@
...
 namespace SuperiorMySqlpp
 {
     namespace Sql
     {
-        using TinyInt = signed char;
-        using SmallInt = short int;
-        using MediumInt = int;
-        using Int = int;
-        using BigInt = long long int;
+        using TinyInt = int_fast8_t;
```
majcl: I don't think it's good idea. These types can be of arbitrary size. And it forces you to use same types in your app. Moreover, possible serialization of these types can be non-portable.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -60,7 +61,7 @@ namespace SuperiorMySqlpp
         constexpr inline std::enable_if_t<std::is_integral<T>::value>
         initializeParamBinding(MYSQL_BIND& binding, T& value)
         {
-            using PureType_t = std::decay_t<T>;
+            using PureType_t = to_fixed_width_int_t<std::decay_t<T>>;
```
majcl: I don't see why?

majcl (changes requested):

include/superior_mysqlpp/sql_types.hpp
OpatrilPeter: Good idea compared to what? You are right that these definitions are not acceptable for serialization, but they do not try to be - they are exactly comparable to current definitions. For example - `int` in C is defined with identical constraints as `int_fast16_t`, but is way less clear about those. Take `MediumInt` - this one has 24bits, so on architecture where int is 16b only current definition will break.

Alternative is using exact sized types such as `int16_t` - but that is not entirely portable. Some architectures may not support some of them and there is no int24_t anyway (if the goal is optimality in serialization). Notably, exact types/`least` sized types are somewhere suboptimal speedwise.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: I don't remember this context exactly, but the reason for `to_fixed_width_int_t` in general is to deal with the fact that DB type is defined to be, say, 64b - so on your architecture you can use `long` and `long long` - however, those types are not equivalent for purposes of `std::is_same` or similar - so this normalizes types to matching fixed width.

include/superior_mysqlpp/sql_types.hpp
majcl: C API sdk defines exactly, what mysql and C-types equivalents. See
https://dev.mysql.com/doc/refman/8.0/en/c-api-prepared-statement-type-codes.html

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
majcl: Same case as above.

include/superior_mysqlpp/sql_types.hpp
OpatrilPeter: Excellent reference, thanks for bringing it up. I was under false impression that MySQL library we use does not specify it (as is the case in many other details).
In this light I'd say this PR is not needed, I'll update it to just contain the reference above in form of a comment.

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: Right, the ambiguity I mentioned is cleanly resolved with the mentioned reference.
One can argue that users depending on interface of underlying C library is violating our encapsulation, but it's only theoretical issue.

smradCZ (approved):

majcl (approved):

PR #71, opened 2018-03-02
Minor bugfixes

OpatrilPeter:
Fixes several cases where behaviour was determined to be obviously incorrect.

### DIFF
--- a/include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
+++ b/include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
@@ -46,10 +46,10 @@ namespace SuperiorMySqlpp
         template<typename T>
         constexpr inline void initializeNullable(MYSQL_BIND& binding, Nullable<T>& nullable)
         {
-            // Checks if pointer of C client pointer to their custom bool bool variant is compatible with bool*
-            // TODO: It would be wise to check also sizeof(my_bool) == sizeof(bool)
             static_assert(sizeof(decltype(binding.is_null)) == sizeof(decltype(&nullable.detail_getNullRef())),
                 "Pointers to null indicators must have same size.");
+            static_assert(sizeof(decltype(*binding.is_null)) == sizeof(decltype(nullable.detail_getNullRef())),
+                "Representations of boolean null indicators must be equivalent in SuperiorMysqlpp and C client backend.");
             binding.is_null = reinterpret_cast<my_bool*>(&nullable.detail_getNullRef());
         }
 
--- a/include/superior_mysqlpp/types/nullable.hpp
+++ b/include/superior_mysqlpp/types/nullable.hpp
@@ -371,7 +371,7 @@ public:
         if (engaged && other.engaged)
         {
             ::std::swap(payload, other.payload);
-            ::std::swap(null, null);
+            ::std::swap(null, other.null);
         }
         else if (engaged)
         {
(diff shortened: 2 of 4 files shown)


include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -47,9 +47,9 @@ namespace SuperiorMySqlpp
         constexpr inline void initializeNullable(MYSQL_BIND& binding, Nullable<T>& nullable)
         {
             // Checks if pointer of C client pointer to their custom bool bool variant is compatible with bool*
-            // TODO: It would be wise to check also sizeof(my_bool) == sizeof(bool)
-            static_assert(sizeof(decltype(binding.is_null)) == sizeof(decltype(&nullable.detail_getNullRef())),
-                "Pointers to null indicators must have same size.");
+            static_assert(sizeof(decltype(binding.is_null)) == sizeof(decltype(&nullable.detail_getNullRef()))
+                && sizeof(decltype(*binding.is_null)) == sizeof(decltype(nullable.detail_getNullRef())),
```
majcl: Is it really needed? Nullable flag is initializable even without second condition, isnt it?

include/superior_mysqlpp/types/nullable.hpp
```diff
@@ -371,7 +371,7 @@ class Nullable : public detail::NullableBase
         if (engaged && other.engaged)
         {
             ::std::swap(payload, other.payload);
-            ::std::swap(null, null);
+            ::std::swap(null, other.null);
```
majcl, line 374: ups :-O

majcl (changes requested):

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: `binding.is_null` has type `my_bool*`, &nullable.detail_getNullRef() has type `bool*`, we feed our pointer to C client.
So, we clearly need sanity checks that `my_bool` has same size as `bool`, the second condition. That said, we perhaps do NOT need the first one, as the second one likely implies the first one, that their respective pointers have matching sizes. Opinions?

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
```diff
@@ -47,9 +47,9 @@ namespace SuperiorMySqlpp
         constexpr inline void initializeNullable(MYSQL_BIND& binding, Nullable<T>& nullable)
         {
             // Checks if pointer of C client pointer to their custom bool bool variant is compatible with bool*
```
smradCZ: Doubled `bool bool`

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
smradCZ: I'd prefer remove the first part of condition, checking pointer sizes. The second condition make sense to me.

smradCZ (changes requested):

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: Oh yes, the `bool bool` type, you must have heard about it. Just like `long long` means "it's very long", it's meant to be used only when your bool is really booleany, no mere disguised integer ... that, or I just made a typo. 😛

smradCZ (dismissed):

include/superior_mysqlpp/prepared_statements/initialize_bindings.hpp
OpatrilPeter: After further offline discussion, we ended up keeping both conditions, main argument being that the original condition was done by original lib author, although without commentary what real situation it was meant to cover.

majcl (approved):

PR #75, opened 2018-03-16
Dynamic types support

MrWarlockX:
Adds support for dynamically-allocated strings and blobs


majcl:
Actually now, library now provides Dynamic<T> as new data type. Is it possible to use it as Param?

OpatrilPeter:
This PR is quite old and was stuck in limbo due to internal disagreements how to handle this functionality.
The discussions came to a halt, as this feature wasn't actually considered significant enough to any stakeholder.
As the member responsible for this PR is not working on our codebase anymore and there's still no drive behind it years later, I think it'll be best to close this one and possibly revisit the idea later.

PR #78, opened 2018-03-29
Documentation of DBDriver and mysql_hacks in low-level module

smradCZ:


include/superior_mysqlpp/low_level/mysql_hacks.hpp
```diff
@@ -75,6 +79,8 @@ namespace SuperiorMySqlpp { namespace LowLevel { namespace detail { namespace My
 {
     /*
      * Use this class only with thread_local storage specifier to ensure correct functionality!
+     *
+     * RAII class
```
majcl: Please add maybe short description or example, where and why it is used.

majcl (changes requested):
Also There is a lot of not noexcept functions, that doesn't have throws section. Probably a some noexcept are missing. Please create an issue for this.

include/superior_mysqlpp/low_level/mysql_hacks.hpp
smradCZ: Issue was filled, see #83 

Class comment was updated, please check updated PR.

majcl (approved):

PR #96, opened 2018-08-16
Refactor and document dnsa pool.

AdamStepan:


include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -193,6 +228,9 @@ namespace SuperiorMySqlpp
                 enabled = false;
```
MrWarlockX, line 244: This seems to me like duplicit operation (if condition *jobThread.joinable()* passes)

AdamStepan: You are right, it should be in "else" branch.

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -201,21 +239,34 @@ namespace SuperiorMySqlpp
                 jobThread = std::thread{&DnsAwarePoolManagement::job, std::cref(*this)};
             }
 
+            /**
+             * Returns true if jobThread is in joinable state.
+             */
             auto isDnsAwarePoolManagementThreadRunning() const
             {
+                // XXX: this is ambigious, it can return false if jobThread is detached
```
OpatrilPeter: Your point is correct, however  `jobThread` is private and I can't see any detaching - so it's more like a note about possible danger? I'd reformulate it that way.

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -201,21 +239,34 @@ namespace SuperiorMySqlpp
                 jobThread = std::thread{&DnsAwarePoolManagement::job, std::cref(*this)};
             }
 
+            /**
+             * Returns true if jobThread is in joinable state.
```
OpatrilPeter: IMHO point of this function is checking if management thread is running - how it does so is implementation detail. See comment below on discussion about implementation.

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -63,15 +72,23 @@ namespace SuperiorMySqlpp
             }
         };
 
-
+        /**
+         * Mix in class that checks and logs DNS changes.
```
OpatrilPeter: Mixin

include/superior_mysqlpp/dnsa_connection_pool.hpp
OpatrilPeter: `ambiguous`

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,99 @@ namespace SuperiorMySqlpp
...
             {
                 return *static_cast<const Base*>(this);
             }
 
-            /*
+            /**
+             * This function is called when unexpected error is thrown during dns resolution error handling.
+             * It's called only if logger throws exception.
+             */
+            void onUnexpectedError(void) const noexcept {
```
OpatrilPeter: Please use bracketing consistent with the rest of the code (that includes `catch` etc.)

AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,99 @@ namespace SuperiorMySqlpp
...
+                }
+            }
+
+            /**
              * This function is run as parallel thread.
              * We cannot afford to let any exceptions to be propagated from logging functions
-             * since it would result in call to std::terminate()!!! Hence all the try/catch blocks.
+             * since it would result in call to std::terminate().
+             */
+            void job(void) const {
```
OpatrilPeter: Explicit void in parameters is completely redundant in C++ (not in C)

AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -39,6 +41,13 @@ namespace SuperiorMySqlpp
             DnsResolver& operator=(const DnsResolver&) = default;
             DnsResolver& operator=(DnsResolver&&) = default;
 
+            /**
+             * Resolve a hostname to ip addresses.
+             *
+             * @tparam const reference to std::string
+             * @param hostname
+             * @return Vector of strings that represent ip addresses for given hostname.
+             */
```
majcl: I'm wondering if it wouldn't be nice whether methods docs should have throws section. 
At least here, IMHO it makes sense to mention possible asio exceptions.

majcl: What do you think?
Also @OpatrilPeter @smradCZ

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -81,6 +98,10 @@ namespace SuperiorMySqlpp
 
             DnsAwarePoolManagement(const DnsAwarePoolManagement&) = delete;
 
+            /**
+             * Initializes pool from other pool.
+             * It starts DNS resolution job if job is running in other instance.
```
majcl, line 109: 👍

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,106 @@ namespace SuperiorMySqlpp
...
             }
 
-            /*
+            /**
+             * This function is called when unexpected error is thrown during dns resolution error handling.
+             * It's called only if logger throws exception.
+             */
+            void onUnexpectedError() const noexcept
+            {
+                if(terminateOnFailure) {
```
majcl: space after if

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,106 @@ namespace SuperiorMySqlpp
...
-                        std::terminate();
-                    }
-                };
-
                 try
                 {
-                    std::vector<std::string> lastIpAddresses{};
+                    runJobLoop();
+                }
+                catch(std::exception &e)
```
majcl: Also space after catches.

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,106 @@ namespace SuperiorMySqlpp
...
+                if(terminateOnFailure) {
+                    std::terminate();
+                }
+            }
+
+            /**
              * This function is run as parallel thread.
              * We cannot afford to let any exceptions to be propagated from logging functions
-             * since it would result in call to std::terminate()!!! Hence all the try/catch blocks.
+             * since it would result in call to std::terminate().
```
majcl, line 164: Right, but in outer try/catch block, we also try to log -> so the problem is still there, isn't it?

include/superior_mysqlpp/dnsa_connection_pool.hpp
```diff
@@ -101,88 +125,106 @@ namespace SuperiorMySqlpp
...
-                    getBase().getLogger()->logSharedPtrPoolDnsAwarePoolManagementError(getBase().getId(), e);
-                    onError();
-                }
-                catch (...)
-                {
-                    getBase().getLogger()->logSharedPtrPoolDnsAwarePoolManagementError(getBase().getId());
-                    onError();
+                    getBase().getLogger()->logSharedPtrPoolDnsAwarePoolManagementCycleEnd(getBase().getId());
+
+                    sleepInParts(sleepTime, std::chrono::milliseconds{50}, [&](){ return enabled.load(); });
```
majcl: I think that 50ms should probably be at least some kind of "explicit" constant?

majcl (changes requested):

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: ACK

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: OK, I'll to document all exceptions which are thrown by asio.

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: Yes, the problem is still there. I will remove logger's logging from the outer catch block.

include/superior_mysqlpp/dnsa_connection_pool.hpp
AdamStepan: What about "shutdownCheckPeriod"?

PR #97, opened 2018-08-16
Add support for std::string to makeHexString

### LINKED ISSUE #79: makeHexString suitable for std::string
Method `makeHexString` is usable only for `char[]` (and maybe custom types supporting std::begin, std::end and default conversion to char[]) but not for C-string represented as `const char *`, for std::string c_str() method neither.

Also using `std::distance` for getting item count "Iterable" type should be better than getting it by difference of end and begin iterator.

AdamStepan:
It closes #79

### DIFF
--- a/include/superior_mysqlpp/low_level/dbdriver.hpp
+++ b/include/superior_mysqlpp/low_level/dbdriver.hpp
@@ -18,6 +18,7 @@
 #include <type_traits>
 #include <cinttypes>
 #include <atomic>
+#include <cstring>
 
 #include <superior_mysqlpp/logging.hpp>
 #include <superior_mysqlpp/exceptions.hpp>
@@ -319,23 +320,50 @@ namespace SuperiorMySqlpp { namespace LowLevel
             return escapeStringNoConnection(original.c_str(), original.length());
         }
 
+        // XXX: makeHexString functions can be implemented by one function using constexpr-if from c++17
         /**
          * Converts string to HEX string suitable for using in SQL statements.
          * Useful for binary data for instance.
          * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
          *
-         * @tparam Type supporting std::begin, std::end and default conversion to char array (char[]).
+         * @param from Original string.
+         * @param fromlen Length of original string.
+         * @return HEX string.
+         */
+        std::string makeHexString(const char *from, std::size_t fromlen)
+        {
+            std::string result(fromlen * 2 + 1, '\0');
+
+            auto length = mysql_hex_string(&result.front(), from, fromlen);
+            result.resize(length);
+
+            return result;
+        }
+
+        /**
+         * Converts string to HEX string suitable for using in SQL statements.
+         * Useful for binary data for instance.
+         * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
+         *
          * @param from Original string.
          * @return HEX string.
          */
-        template<typename Iterable>
-        std::string makeHexString(const Iterable& from)
+        std::string makeHexString(const char *from)
         {
-            auto fromSize = std::end(from) - std::begin(from);
-            auto buffer = std::make_unique<char[]>(fromSize*2 + 1);
-            auto length = mysql_hex_string(buffer.get(), from, fromSize);
+            return makeHexString(from, std::strlen(from));
+        }
 
-            return std::string(buffer.get(), length);
+        /**
+         * Converts string to HEX string suitable for using in SQL statements.
+         * Useful for binary data for instance.
+         * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
+         *
+         * @param from Original string.
+         * @return HEX string.
+         */
+        std::string makeHexString(const std::string &from)
+        {
+            return makeHexString(from.data(), from.size());
         }
 
         /**
--- a/packages/_debian-common/changelog
+++ b/packages/_debian-common/changelog
@@ -1,4 +1,4 @@
-libsuperiormysqlpp (0.3.4) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
 
   [ Radek Smejdir ]
   * Add ConnectionConfiguration
@@ -11,8 +11,9 @@ libsuperiormysqlpp (0.3.4) UNRELEASED; urgency=medium
   * Propagate return values from mysql_*row_seek methods
   * Fix dnsa test
   * Add noexcept qualifier to database driver methods
+  * Fix behavior of makeHexString
 
- -- Adam Stepan <adam.stepan@firma.seznam.cz>  Thu, 16 Aug 2018 09:19:27 +0200
+ -- Adam Stepan <adam.stepan@firma.seznam.cz>  Thu, 16 Aug 2018 09:09:37 +0200
 
 libsuperiormysqlpp (0.3.2) unstable; urgency=medium
 
--- a/tests/db_access/driver.cpp
+++ b/tests/db_access/driver.cpp
@@ -13,7 +13,7 @@
 using namespace bandit;
 using namespace SuperiorMySqlpp;
 using namespace SuperiorMySqlpp::LowLevel;
-
+using namespace std::string_literals;
 
 go_bandit([](){
     describe("Test driver", [&](){
@@ -22,6 +22,14 @@ go_bandit([](){
         it("will initialize", [&](){
             DBDriver driver{};
             driver.connect(s.host.c_str(), s.user.c_str(), s.password.c_str(), nullptr, s.port, nullptr);
+        });
+
+        it("can make hex string from string types", [&]() {
+            DBDriver driver{};
+            std::string expected{"666F6F626172"};
+
+            AssertThat(driver.makeHexString("foobar"), Equals(expected));
+            AssertThat(driver.makeHexString("foobar"s), Equals(expected));
         });
     });
 });


include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,6 +319,28 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
+         * @param from Original string.
+         * @tparam Size type.
+         * @param length of original string
+         * @return HEX string.
+         */
+        template <typename T>
+        std::string makeHexString(T *form, std::size_t formlen)
+        {
+
+            auto buffer = std::make_unique<char[]>(formlen * 2 + 1);
```
MrWarlockX: I would rather create std::string, resize it, use it as a buffer and then return it directly to the caller (there will be only one allocation instead of two)

packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) unstable; urgency=medium
```
MrWarlockX: I would leave UNRELEASED here, if you take look at *master* commits, you will find release commits, that changes this line.

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -331,11 +353,24 @@ namespace SuperiorMySqlpp { namespace LowLevel
         template<typename Iterable>
         std::string makeHexString(const Iterable& from)
         {
-            auto fromSize = std::end(from) - std::begin(from);
-            auto buffer = std::make_unique<char[]>(fromSize*2 + 1);
-            auto length = mysql_hex_string(buffer.get(), from, fromSize);
+            return makeHexString(from, std::end(from) - std::begin(from));
```
MrWarlockX: Are you sure about if the first argument is correct? You call `std::string makeHexString(T *form, std::size_t formlen)` function, where first argument is pointer and you are currently passing reference.

tests/db_access/driver.cpp
```diff
@@ -23,6 +24,19 @@ go_bandit([](){
...
+
+        it("can make hex string from string types", [&]() {
+            DBDriver driver{};
+            std::string expected{"666F6F62617200"};
+
+            AssertThat(driver.makeHexString("foobar"), Equals(expected));
+            AssertThat(driver.makeHexString("foobar"s), Equals(expected));
+
+            std::string foobar{"foobar"};
+            AssertThat(driver.makeHexString(foobar), Equals(expected));
```
MrWarlockX: Isn't this test already on line 33? :)

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: Well, it's expected that first argument will be convertible to pointer type (char[] for instance), in that case argument will be type of const reference to pointer.

tests/db_access/driver.cpp
AdamStepan: Nope, assert on line 33 calls method with rvalue, this assert calls method with lvalue.

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: You are right, I actually tried it, but it works only when I used this ugly hack with -1.
```C++
    std::string result(formlen * 2 + 1, '\0');

    auto length = mysql_hex_string(const_cast<char *>(result.data() - 1), form, formlen);
    result.resize(length);

    return result;

```
If I don't use -1 first character of string remains the same.
I didn't want to use this version because it's odd.

packages/_debian-common/changelog

include/superior_mysqlpp/low_level/dbdriver.hpp
MrWarlockX: Did you tried to use `&*result.begin()`? And then, you don't need that extra null byte, it is there by default.

```c++
std::string result;
result.resize(formlen * 2);
result.resize(mysql_hex_string(&*result.begin(), form, formlen));
return result;
```

tests/db_access/driver.cpp
MrWarlockX: Ok, however after the last change *makeHexString()* takes const lvalue, thus it does the same thing and so this is redundant test :) .

AdamStepan: ACK

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,23 +320,54 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
+
+            return result;
+        }
+
+        /**
+         * Converts string to HEX string suitable for using in SQL statements.
+         * Useful for binary data for instance.
+         * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
+         *
+         * @tparam Type const char *
```
smradCZ: There is no template param

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,23 +320,54 @@ namespace SuperiorMySqlpp { namespace LowLevel
             return escapeStringNoConnection(original.c_str(), original.length());
         }
 
+        // XXX: makeHexString functions can be implemented by one function using constexpr-if from c++17
```
smradCZ, line 323: Please rename XXX to TODO or something more describing the purpose of the comment

packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) UNRELEASED; urgency=medium
```
smradCZ: You don't need to increment patch version because this version (0.3.3) is not released yet.

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: ACK

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: XXX is IMHO quite standard, it can be found in Linux source code, Python standard library or GNU C++ library.

packages/_debian-common/changelog

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,23 +320,53 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
         /**
          * Converts string to HEX string suitable for using in SQL statements.
          * Useful for binary data for instance.
          * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
          *
-         * @tparam Type supporting std::begin, std::end and default conversion to char array (char[]).
+         * @tparam Type convertable to char[]
```
smradCZ: convertible

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,23 +320,53 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
+        {
+            return makeHexString(from, std::strlen(from));
+        }
+
+        /**
+         * Converts string to HEX string suitable for using in SQL statements.
+         * Useful for binary data for instance.
+         * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
+         *
+         * @tparam Type imlicitly convertable to std::string.
```
smradCZ: implicitly convertible

smradCZ: but the method have no template params at all

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: ACK

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: ACK

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -319,23 +320,50 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
 
+        // XXX: makeHexString functions can be implemented by one function using constexpr-if from c++17
         /**
          * Converts string to HEX string suitable for using in SQL statements.
          * Useful for binary data for instance.
          * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-hex-string.html
          *
-         * @tparam Type supporting std::begin, std::end and default conversion to char array (char[]).
          * @param from Original string.
+         * @param length of original string
```
smradCZ: missing param name `fromLen`

tests/db_access/driver.cpp
```diff
@@ -23,6 +23,16 @@ go_bandit([](){
...
         });
+
+        it("can make hex string from string types", [&]() {
+            DBDriver driver{};
+            std::string expected{"666F6F626172"};
+
+            AssertThat(driver.makeHexString("foobar"), Equals(expected));
+            AssertThat(driver.makeHexString("foobar"s), Equals(expected));
+        });
+
```
smradCZ: Please remove following empty lines.

include/superior_mysqlpp/low_level/dbdriver.hpp
smradCZ: Also please write it as a sentence - First letter in uppercase and dot on its end.

smradCZ (changes requested):

tests/db_access/driver.cpp
AdamStepan: ACK

include/superior_mysqlpp/low_level/dbdriver.hpp
AdamStepan: ACK

smradCZ (dismissed):

smradCZ (approved):

PR #98, opened 2018-08-16
Propagate return value of mysql_*row_seek methods

### LINKED ISSUE #81: Propagate return value of mysql_row_seek and mysql_stmt_row_seek
Methods `DBDriver::seekRowOffset` and `DBDriver::Statement::seekRowOffset` should return previous row offset. The offset can be retrieved directly from return value of called underlying library functions `mysql_row_seek` or `myql_stmt_row_seek`.

AdamStepan:
It closes #81

### DIFF
--- a/packages/_debian-common/changelog
+++ b/packages/_debian-common/changelog
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) unstable; urgency=medium
 
   [ Radek Smejdir ]
   * Add ConnectionConfiguration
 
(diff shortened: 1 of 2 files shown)


packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) unstable; urgency=medium
```
MrWarlockX: I would leave UNRELEASED here, if you take look at master commits, you will find release commits, that changes this line.

majcl (approved):

PR #99, opened 2018-08-16
Add noexcept qualifier to database driver methods.

### LINKED ISSUE #83: DBDriver - many methods can be noexcept
Many methods in DBDriver class can be declared with `noexcept` specifier.
They call underlying C API functions directly with no other logic thus exceptions shouldn't arise there in most cases.

AdamStepan:
It closes #83

### DIFF
--- a/include/superior_mysqlpp/low_level/dbdriver.hpp
+++ b/include/superior_mysqlpp/low_level/dbdriver.hpp
@@ -221,7 +221,7 @@ namespace SuperiorMySqlpp { namespace LowLevel
          * @param value Logger instance.
          */
         template<typename T>
-        void setLoggerPtr(T&& value) noexcept
+        void setLoggerPtr(T&& value)
         {
             loggerPtr = std::forward<T>(value);
         }
--- a/include/superior_mysqlpp/low_level/dbdriver.hpp
+++ b/include/superior_mysqlpp/low_level/dbdriver.hpp
@@ -113,7 +113,7 @@ namespace SuperiorMySqlpp { namespace LowLevel
          * Internal thread-safe ID counter.
          * @return ID counter reference.
          */
-        static auto& getGlobalIdRef()
+        static auto& getGlobalIdRef() noexcept
         {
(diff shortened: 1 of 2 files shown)


packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) unstable; urgency=medium
```
MrWarlockX: I would leave UNRELEASED here, if you take look at master commits, you will find release commits, that changes this line.

packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) UNRELEASED; urgency=medium
```
OpatrilPeter: Why are you incrementing the number if 0.3.3 is unreleased?

OpatrilPeter (changes requested):

smradCZ:
Please rebase this branch onto upstream `devel`.

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -221,7 +221,7 @@ namespace SuperiorMySqlpp { namespace LowLevel
          * @param value Logger instance.
          */
         template<typename T>
-        void setLoggerPtr(T&& value)
+        void setLoggerPtr(T&& value) noexcept
```
MrWarlockX: I think, that noexcept is not appropriate here, because `loggerPtr` is of type `Loggers::SharedPointer_t` which is `std::shared_ptr`. And because this function is templated, you can pass in e.g. `std::unique_ptr`, and in that case it is not noexcept.

AdamStepan: ACK

smradCZ (approved):

Issue #80, opened 2018-04-05
DBDriver::Result::freeResult() is useless


smradCZ:
I think the method DBDriver::Result::freeResult() should be removed. It calls `mysql_free_result` but it will be done automatically in class destructor.

In case the method is really wanted, then some things should be updated:
- it should return `void` instead of magic `auto`
- it should set `resultPtr` to `nullptr` after calling `mysql_free_result` method to ensure destructor won't freeing memory on same pointer again.

majcl:
Merged into devel by pull request #101

PR #101, opened 2018-08-17
Remove useless method DBDriver::Result::freeResult()

tomasprochazka2:
The method DBDriver::Result::freeResult() is never used and thus can be removed

Closes the issue #80

### DIFF
--- a/include/superior_mysqlpp/low_level/dbdriver.hpp
+++ b/include/superior_mysqlpp/low_level/dbdriver.hpp
@@ -447,18 +447,6 @@ namespace SuperiorMySqlpp { namespace LowLevel
             /** Pointer to underlying library's result set instance. */
             MYSQL_RES* resultPtr;
 
-            /**
-             * Frees allocated memory for current result set.
-             * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-free-result.html
-             */
-            void close() noexcept
-            {
-                if (resultPtr != nullptr)
-                {
-                    mysql_free_result(resultPtr);
-                }
-            }
-
         public:
             /**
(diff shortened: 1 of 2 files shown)


include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -627,18 +627,7 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
-             * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-free-result.html
-             *
-             * @return Void - nothing!
-             */
-            auto freeResult()
-            {
-                return mysql_free_result(resultPtr);
-            }
-
+
```
MrWarlockX: Please remove unnecessary whitespaces.

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -628,17 +628,6 @@ namespace SuperiorMySqlpp { namespace LowLevel
                 return mysql_field_tell(resultPtr);
             }
 
-            /**
-             * Frees allocated memory for current result set.
-             * @see https://dev.mysql.com/doc/refman/5.7/en/mysql-free-result.html
-             *
-             * @return Void - nothing!
-             */
-            auto freeResult()
```
majcl: IMHO would be better to leave it here, but call `mysql_free_result` with same condition as in close(), make it noexcept, and possibly remove close() method and instead use this one? @smradCZ ?

majcl (changes requested):

majcl (approved):

PR #102, opened 2018-08-20
Fix dnsa test

AdamStepan:


packages/_debian-common/changelog
```diff
@@ -1,9 +1,12 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.3.4) UNRELEASED; urgency=medium
```
smradCZ: You don't need to raise patch version, previous changelog is still UNRELEASED

majcl (approved):

PR #105, opened 2019-01-24
psReadValues: Extend exceptions

MrWarlockX:
Extends exceptions in `psReadValues` function.

### DIFF
--- a/include/superior_mysqlpp/exceptions.hpp
+++ b/include/superior_mysqlpp/exceptions.hpp
@@ -108,7 +108,7 @@ namespace SuperiorMySqlpp
     };
 
     /**
-     * @brief Error for unexpected row count
+     * @brief Error for unexpected row count, accessible via `getRowCount()` method
      */
     class UnexpectedRowCountError : public SuperiorMySqlppError
     {
--- a/include/superior_mysqlpp/exceptions.hpp
+++ b/include/superior_mysqlpp/exceptions.hpp
@@ -114,7 +114,7 @@ namespace SuperiorMySqlpp
         size_t numRows_;
     public:
         UnexpectedRowsCountError(const std::string &message, size_t numRows)
-            : SuperiorMySqlppError { messge }, numRows_(numRows) {}
+            : SuperiorMySqlppError { message }, numRows_(numRows) {}
 
         inline size_t numRows() const
         {
--- a/include/superior_mysqlpp/exceptions.hpp
+++ b/include/superior_mysqlpp/exceptions.hpp
@@ -111,25 +111,15 @@ namespace SuperiorMySqlpp
      * @brief Base class for multiple rows error
      */
     class UnexpectedRowsCountError : public SuperiorMySqlppError {
+        size_t numRows_;
     public:
-        using SuperiorMySqlppError::SuperiorMySqlppError;
+        UnexpectedRowsCountError(const std::string &message, size_t numRows)
+            : SuperiorMySqlppError { messge }, numRows_(numRows) {}
+
+        inline size_t numRows() const
+        {
+            return numRows_;
+        }
     };
 
-    /**
-     * @brief This exception is thrown, when you attempt to read more rows in function `psReadValues`
-     */
-    class UnexpectedMultipleRowsError : public UnexpectedRowsCountError
-    {
-    public:
-        using UnexpectedRowsCountError::UnexpectedRowsCountError;
-    };
-
-    /**
-     * @brief Error when no results are returned
-     */
-    class UnexpectedZeroRowsError : public UnexpectedRowsCountError
-    {
-    public:
-        using UnexpectedRowsCountError::UnexpectedRowsCountError;
-    };
 }
--- a/include/superior_mysqlpp/exceptions.hpp
+++ b/include/superior_mysqlpp/exceptions.hpp
@@ -108,7 +108,7 @@ namespace SuperiorMySqlpp
     };
 
     /**
-     * @brief Error for unexpected row count, accessible via `getRowCount()` method
+     * @brief Error for unexpected row count. Number of rows accessible via `getRowCount()` method
      */
     class UnexpectedRowCountError : public SuperiorMySqlppError
     {
--- a/include/superior_mysqlpp/exceptions.hpp
+++ b/include/superior_mysqlpp/exceptions.hpp
@@ -110,16 +110,19 @@ namespace SuperiorMySqlpp
     /**
      * @brief Error for unexpected row count
      */
-    class UnexpectedRowCountError : public SuperiorMySqlppError {
-        size_t numRows;
+    class UnexpectedRowCountError : public SuperiorMySqlppError
+    {
     public:
-        UnexpectedRowCountError(const std::string &message, size_t numRows)
-            : SuperiorMySqlppError { message }, numRows(numRows) {}
+        UnexpectedRowCountError(const std::string &message, size_t rowCount)
+            : SuperiorMySqlppError { message }, rowCount(rowCount) {}
 
-        inline size_t getNumRows() const
+        inline size_t getRowCount() const
         {
-            return numRows;
+            return rowCount;
         }
+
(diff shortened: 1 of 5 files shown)


include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -140,15 +140,19 @@ namespace SuperiorMySqlpp
...
     void psReadValues(PreparedStatementType &&ps, Args&... values)
     {
         ps.execute();
 
-        if(ps.getRowsCount() != 1 || !ps.fetch())
+        if (ps.getRowsCount() != 1 || !ps.fetch())
         {
+            if (ps.getRowsCount() == 0) {
+                throw UnexpectedZeroRowsError("psReadValues() has to have a single-row result!");
+            }
```
OpatrilPeter: I'd prefer `else` to highlight that the throw is there for negated condition.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -140,15 +140,19 @@ namespace SuperiorMySqlpp
      * @param ps Prepared statement object (only static statements are currently supported)
      * @param values References to variables to be loaded from query
      *               Their type must be compatible with query result types
-     * @throws std::runtime_error if more than one row is being loaded
+     * @throws UnexpectedZeroRowsError if result is empty
+     * @throws UnexpectedMultipleRowserror if more than one row is being loaded
```
OpatrilPeter: Uppercase E in `error`.

OpatrilPeter (dismissed):
Current version seems OK to me, as long as commits are squashed before merge.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -140,16 +140,16 @@ namespace SuperiorMySqlpp
...
     template<typename PreparedStatementType, typename... Args, typename = typename std::enable_if<std::is_base_of<detail::StatementBase, std::remove_reference_t<PreparedStatementType>>::value>::type>
     void psReadValues(PreparedStatementType &&ps, Args&... values)
     {
         ps.execute();
 
-        if(ps.getRowsCount() != 1 || !ps.fetch())
+        if (ps.getRowsCount() != 1 || !ps.fetch())
         {
-            throw UnexpectedMultipleRowsError("psReadValues() can read only one row per query!");
+            throw UnexpectedRowCountError("psReadValues() must read exactly one row", ps.getRowsCount());
```
majcl: I would reword it slightly. Actually psReadValuues() can still behave as it would read exactly one row. We should mention, that provided query/prepared statement read more than one line

majcl (changes requested):

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
majcl: I don't get it... can you explain by example?

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -108,11 +108,25 @@ namespace SuperiorMySqlpp
     };
 
     /**
-     * @brief This exception is thrown, when you attempt to read more rows in function `psReadValues`
+     * @brief Error for unexpected row count
```
majcl: There should be more thorough explanation. You should mention, what additional information exception provides.

majcl:
Readme.md should mention this change. With api-changes warning .

tests/db_access/prepared_statements.cpp
```diff
@@ -603,7 +603,7 @@ go_bandit([](){
         it("can work with psReadValues (throws multiple rows error)", [&](){
             int id{};
             BlobData blob{}, binary{}, varbinary{};
-            AssertThrows(UnexpectedMultipleRowsError,
+            AssertThrows(UnexpectedRowCountError,
```
majcl, line 617: Both scenarios should be tested. 0 returned rows, and multiple returned rows.

packages/_debian-common/changelog
```diff
@@ -1,4 +1,4 @@
-libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
+libsuperiormysqlpp (0.4.0) UNRELEASED; urgency=medium
```
majcl, line 1: I'm wondering, whether we should mention api breaking changes within changelog. Normally these changes would force major version update, but as we are still in version 0, this is a little bit different situation. But maybe explicit API changes warning could save someone headaches.

majcl (changes requested):

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
OpatrilPeter: Put the other throw in else branch, which by now is already done.

tests/db_access/prepared_statements.cpp
```diff
@@ -600,10 +600,18 @@ go_bandit([](){
             AssertThat(varbinary.getStringView(), Equals("ij\0kl"s));
         });
 
-        it("can work with psReadValues (throws multiple rows error)", [&](){
+        it("can work with psReadValues (zero rows error)", [&](){
             int id{};
             BlobData blob{}, binary{}, varbinary{};
```
smradCZ: I prefer made tests too much simple as possible - for instance, select varies of parameters has no added value for this test-case, the ORDERing the results in a query is useless as well.

This test-case may be reduced to form like this:
```suggestion
            int id{};
            AssertThrows(UnexpectedRowCountError,
                psReadValues("SELECT `id` FROM `test_superior_sqlpp`.`binary_data` LIMIT 0", connection, id)
            );
```

The same for `multiple rows error` test case below.

smradCZ (changes requested):

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -108,11 +108,25 @@ namespace SuperiorMySqlpp
     };
 
     /**
-     * @brief This exception is thrown, when you attempt to read more rows in function `psReadValues`
+     * @brief Error for unexpected row count, accessible via `getRowCount()` method
```
majcl: This is really strange wording. Please describe exceptions more carefully.

packages/_debian-common/changelog
```diff
@@ -14,6 +14,10 @@ libsuperiormysqlpp (0.3.3) UNRELEASED; urgency=medium
   * Add noexcept qualifier to database driver methods
   * Fix behavior of makeHexString
 
+  [ Michal Merc ]
+  * !!! BREAKING API CHANGE !!!
```
majcl: Maybe breaking API changes should be mentioned at the top of changelog?

majcl (changes requested):

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
majcl: But now there is only one exception, isn't it? There is no need for this inner `ps.getRowsCount() == 0` if

OpatrilPeter: Yes, it's not relevant in current version anymore.

smradCZ (dismissed):

tests/db_access/prepared_statements.cpp
```diff
@@ -600,11 +600,17 @@ go_bandit([](){
             AssertThat(varbinary.getStringView(), Equals("ij\0kl"s));
         });
 
-        it("can work with psReadValues (throws multiple rows error)", [&](){
+        it("can work with psReadValues (zero rows error)", [&](){
```
majcl: Shoudln't it say something like psReadValues can detect wrong result? (here it means zero result).
Actually there should probably be `describe section` for `psReadValues`. And `"zero rows error"` should be one of its tests.

majcl (changes requested):

PR #106, opened 2019-02-15
Refactor: using long-opts in build scripts

smradCZ:
Using long variants of parameters is self-describing and it makes build scripts much more clear to reader.
Also `docker run --rm` was used when packages are build instead of calling `docker run` and `docker rm` separately.

### DIFF
--- a/makefile
+++ b/makefile
@@ -17,12 +17,12 @@ libdir =$(prefix)/lib
 includedir =$(prefix)/include
 
 INSTALL :=install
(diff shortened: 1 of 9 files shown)


majcl (dismissed):

majcl:
This actually depends on pull request #108 .

OpatrilPeter (dismissed):

OpatrilPeter (approved):

PR #110, opened 2019-04-08
feat: Detect lost connection when using query

MrWarlockX:


include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -624,9 +624,16 @@ namespace SuperiorMySqlpp { namespace LowLevel
              *
              * @return MYSQL_ROW with next row results. NULL if there are no more rows or on error.
              */
-            auto fetchRow() noexcept
+            auto fetchRow()
```
OpatrilPeter: Is this API breaking change, or is this function strictly private?

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -624,9 +624,16 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
             {
-                return mysql_fetch_row(resultPtr);
+                auto row = mysql_fetch_row(resultPtr);
```
OpatrilPeter: As a rule of thumb, if the type is meant to be treated as a pointer (such as in this function), I'd highlight this fact by writing it as `auto *`.

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -624,9 +624,16 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
+
+                if (row == nullptr && mysql_errno(resultPtr->handle) != 0)
+                {
+                    throw RuntimeError { "Failed to fetch row, probably lost connection to server" };
```
OpatrilPeter: How about using https://dev.mysql.com/doc/refman/5.7/en/mysql-error.html for actual error message?

OpatrilPeter (changes requested):

include/superior_mysqlpp/low_level/dbdriver.hpp
MrWarlockX: Well, it depends. It is low-level driver which shouldn't anyone use and functions using `fetchRow()` are non-noexcept. And there is API breaking change warning in changelog.

majcl: Though it is propagated upwards, isn't it? so even calling function now cannot be `noexcept`, and it was previously, wasn't it?

MrWarlockX: I am affraid that it was not. Method(s) using this function (`QueryResult::fetchRow()`) were not marked as noexcept.

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -624,9 +624,16 @@ namespace SuperiorMySqlpp { namespace LowLevel
              *
              * @return MYSQL_ROW with next row results. NULL if there are no more rows or on error.
              */
-            auto fetchRow() noexcept
+            auto fetchRow()
             {
-                return mysql_fetch_row(resultPtr);
+                auto *row = mysql_fetch_row(resultPtr);
+
+                if (row == nullptr && mysql_errno(resultPtr->handle) != 0)
```
majcl, line 642: Consider using different fetchRow for store vs use fetch. 
Do we really need this condition for store mode?
Is store mode configurable by template? If so, then in store mode this method (and subsequently its caller) can be marked as noexcept.

packages/_debian-common/changelog
```diff
@@ -5,6 +5,7 @@ libsuperiormysqlpp (0.4.0) UNRELEASED; urgency=medium
   [ Michal Merc ]
   * refactor: Remove szn-debian-wheezy support
   * Extend psReadValues exceptions
+  * feat: Detect lost connection when using query
```
majcl, line 8: It's breaking change only in internal. But as the whole library is header only - even implementation details are actually interface. Im not sure if we need to mark it as API breaking change, but at least note in changelog should be made.

majcl (dismissed):

include/superior_mysqlpp/low_level/dbdriver.hpp
MrWarlockX: As far as I know, it is not configurable by template. However this should be trivial to implement.

OpatrilPeter (approved):

packages/_debian-common/changelog
OpatrilPeter: If you expand this view, you'll see there is such warning due to other Michal's changes.

PR #111, opened 2019-05-23
Log errors in default logger + refactor

### LINKED ISSUE #50: Default logger should also log errors to std::cerr
Citation from README.md
> In default configuration it logs only warnings and errors to std::cerr.

But `Default` logger logs only warnings, errors are omitted.

OpatrilPeter:
Some minor formating may still be in order, but the logic should be ok.
Fixes https://github.com/seznam/SuperiorMySqlpp/issues/50

Edit: after first internal review, this should be ready.

### DIFF
--- a/include/superior_mysqlpp/logging.hpp
+++ b/include/superior_mysqlpp/logging.hpp
@@ -12,40 +12,32 @@
 #include <cinttypes>
 #include <atomic>
 #include <mutex>
+#include <utility>
 
 #include <superior_mysqlpp/types/string_view.hpp>
 #include <superior_mysqlpp/types/spin_guard.hpp>
 #include <superior_mysqlpp/prepared_statements/validate_metadata_modes.hpp>
 #include <superior_mysqlpp/prepared_statements/get_binding_type.hpp>
+#include <superior_mysqlpp/utils.hpp>
 
 
 namespace SuperiorMySqlpp {
 
-    namespace detail
-    {
-        template<typename... Args>
-        void dummy(Args&&...) {}
-    }
+    namespace detail {
 
-    template<typename Stream, typename T>
-    auto& stringify_helper(Stream& stream, T&& value)
-    {
-        return stream << value;
-    }
+        /**
+         * Helper function printing its arguments into std::cerr via operator <<
+         * @remark Parameters are perfect forwarded and are followed by implicit std::endl.
+         * @param lock Reference to atomic lock used to gate the output
+         * @param args Variable number of arguments of any type
+         */
+        template <typename... Args>
+        void logStderr(std::atomic_flag &lock, Args&&... args) {
+            SuperiorMySqlpp::SpinGuard guard{lock};
+            (void) guard; // Unused variable otherwise
+            streamify(std::cerr, std::forward<Args>(args)...) << std::endl;
+        }
 
-    template<typename Stream, typename T, typename... Args>
-    void stringify_helper(Stream& stream, T&& value, Args&&... args)
-    {
-        stringify_helper(stream, value);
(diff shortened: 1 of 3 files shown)


include/superior_mysqlpp/logging.hpp
```diff
@@ -63,6 +38,27 @@ namespace SuperiorMySqlpp {
          * so you shall either implement all logging functions or only those you want.
          */
 
+        namespace detail {
+
+            /**
+             * Global variable guaranteeing single access to stderr from standard implementation of loggers
+             * @remark Conceptally should be hidden in #logStderr function, but that would lead to multiple instances.
+             */
+            std::atomic_flag stderrLogWriteLock{};
```
OpatrilPeter: Global variables are usually warning sing, so I'm pointing it out. My previous solution passed `atomic_lock`s around as they were used previously, this one leads to less boilerplate with no obvious disadvantage for me.

OpatrilPeter: My bad, this rewrite looks nicer then original approach, but global variables in header-only lib are no-no, of course.

include/superior_mysqlpp/logging.hpp
```diff
@@ -12,40 +12,31 @@
...
-    template<typename Stream, typename T, typename... Args>
-    void stringify_helper(Stream& stream, T&& value, Args&&... args)
-    {
-        stringify_helper(stream, value);
-        stringify_helper(stream, std::forward<Args>(args)...);
-    }
+        /**
+         * Helper function printing its arguments into std::cerr via operator <<
+         * @remark Parameters are perfect forwarded and are followed by implicit std::endl.
+         * @param args
```
smradCZ: Please fill up all parameters to doxygen comment

include/superior_mysqlpp/utils.hpp
```diff
@@ -54,4 +54,25 @@ namespace SuperiorMySqlpp
         constexpr std::size_t tuple_size = std::tuple_size<Tuple>::value;
         return detail::invokeViaTupleImpl(f, t, std::make_index_sequence<tuple_size>{});
     }
+
+    /**
+     * Writes variable number of arguments into arbitrary std(-like) stream.
```
smradCZ, line 59: It writes *exactly one* argument

include/superior_mysqlpp/utils.hpp
```diff
@@ -54,4 +54,25 @@ namespace SuperiorMySqlpp
...
+     * @param arg Perfectly forwarded argument
+     * @return Stream reference
+     */
+    template <typename Stream, typename Arg>
+    Stream &streamify(Stream &stream, Arg &&arg)
+    {
+        return stream << std::forward<Arg>(arg);
+    }
+
+    /**
```
smradCZ, line 70: Please fill full doxydoc here to keep whole class documented.

smradCZ (changes requested):
Good point, it's much better to have only one definition of log messages on one place.
There are some small mistakes in doxygen comments, please fix them and then we can merge your changes.
Thanks a lot

include/superior_mysqlpp/utils.hpp
OpatrilPeter: This is by design, goal is that overloads should share the documentation. Apparently, `@overload` was needed for that, judging by other examples in this codebase it wasn't clear.

include/superior_mysqlpp/logging.hpp
OpatrilPeter: Well spotted, forgot about lock being added in last refactor.

smradCZ (approved):

PR #112, opened 2019-06-03
Fix compilation with libc++

AdamStepan:


include/superior_mysqlpp/uncaught_exception_counter.hpp
```diff
@@ -4,32 +4,19 @@
...
-    inline int uncaughtExceptions() noexcept
-    {
-        auto* globalsPtr = __cxxabiv1::__cxa_get_globals();
-        auto* globalsEreasedPtr = reinterpret_cast<char*>(globalsPtr);
-        auto* uncaughtExceptionsErasedPtr = globalsEreasedPtr + sizeof(void*);
-        auto* uncaughtExceptionsPtr = reinterpret_cast<int*>(uncaughtExceptionsErasedPtr);
-        return *uncaughtExceptionsPtr;
-    }
-
     class UncaughtExceptionCounter
```
OpatrilPeter, line 26: I'd rename this to something like `MultipleExceptionDetector` + method then doesn't even need a name, perhaps just operator().
Note the name is also in the filename.

include/superior_mysqlpp/uncaught_exception_counter.hpp
```diff
@@ -4,32 +4,19 @@
 
 #pragma once
 
-#include <cxxabi.h>
+#include <exception>
 
 namespace SuperiorMySqlpp
 {
-    inline int uncaughtExceptions() noexcept
```
OpatrilPeter: I didn't realize before this is not a detail/private, so this removal is a breaking change, we must be aware of that, perhaps even keep this method as deprecated for some time. Opinions, @smradCZ ?

packages/_debian-common/changelog
```diff
@@ -1,8 +1,9 @@
 libsuperiormysqlpp (0.5.1) UNRELEASED; urgency=medium
 
-  * 
+  [ Adam Stepan ]
+  * Fix compilation with libc++
```
OpatrilPeter: Please note what actually changed, fix of clang compilation is essentially a sideeffect.

AdamStepan: Yeah, you are right, I forgot to change it when I had rewritten the `UncaughtExceptionCounter` class.

include/superior_mysqlpp/uncaught_exception_counter.hpp
AdamStepan: I removed it because you mentioned that this implementation it's some kind of hack.

I provided you a solution that worked just fine, but you wanted to use something more portable (e.g. `std::current_exception` instead of some function from `abi`). so if you want me to keep this method i will revert changes of `UncaughtExceptionCounter`.

include/superior_mysqlpp/uncaught_exception_counter.hpp
AdamStepan: I think that name should be some predicate (`MultipleExceptionsOccured`, for example) If it should have only `operator()`.

PR #113, opened 2019-06-03
Implement psQuery function, allowing templated queries

MrWarlockX:
This pull request adds functionality for templated queries, so you can supply params to prepared statements with ease.


include/superior_mysqlpp/extras/function_info.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace SuperiorMySqlpp
+{
+    namespace detail
+    {
+        /**
+         * Detects, if all template parameters are references
+         */
+        template<typename...>
+        struct are_arguments_lrefs;
```
OpatrilPeter: Simpler
```c++
template<typename...>
struct are_arguments_lrefs : std::false_type {};

template<typename... Args>
struct are_arguments_lrefs<Args&...> : std::true_type {};
```

include/superior_mysqlpp/extras/function_info.hpp
```diff
@@ -0,0 +1,63 @@
...
+            using is_volatile   = IsVolatile;
+            using arguments     = std::tuple<Args...>;
+            using raw_arguments = std::tuple<std::remove_cv_t<std::remove_reference_t<Args>>...>;
+
+            using arguments_are_lvalue_references = are_arguments_lrefs<Args...>;
+        };
+    }
+
+    /**
+     * Returns information about function
```
OpatrilPeter: Perhaps I'd describe it as 
```
Type trait providing metadata about functions.
Accepts functions, methods and functors (types with operator()).
```

include/superior_mysqlpp/extras/function_info.hpp
```diff
@@ -0,0 +1,63 @@
...
+    struct FunctionInfo<ResultType (Args...)> : detail::FunctionInfoImpl<ResultType, std::false_type, std::false_type, Args...> {};
+
+    template<typename ResultType, typename... Args>
+    struct FunctionInfo<ResultType (*)(Args...)> : detail::FunctionInfoImpl<ResultType, std::false_type, std::false_type, Args...> {};
+
+    template<typename Class, typename ResultType, typename... Args>
+    struct FunctionInfo<ResultType (Class::*)(Args...)> : detail::FunctionInfoImpl<ResultType, std::false_type, std::false_type, Args...> {};
+
+    template<typename Class, typename ResultType, typename... Args>
+    struct FunctionInfo<ResultType (Class::*)(Args...) const> : detail::FunctionInfoImpl<ResultType, std::true_type, std::false_type, Args...> {};
```
OpatrilPeter: Do your overloads handle signatures like `ResultType (Class::*)(Args...) const &`?

include/superior_mysqlpp/utils.hpp
```diff
@@ -75,4 +75,119 @@ namespace SuperiorMySqlpp
...
+        /**
+         * @brief Generic function info obtained from FunctionInfo trait
+         * 
+         * @tparam ResultType type returned by function
+         * @tparam IsConst true_type, if function has been marked const
+         * @tparam IsVolatile true_type, if function has been marked volatile
+         * @tparam Args function arguments
+         */
+        template<typename ResultType, typename IsConst, typename IsVolatile, typename... Args>
+        struct FunctionInfoImpl
```
OpatrilPeter, line 101: This seems like a useful general purpose trait. As such, I'd rather seen him _minimal_ - destructuring the function type into arguments, result type and qualifiers, but nothing more.
Specifically - raw_arguments  + arguments_are_lvalue_references  would work better as separate traits.
That way is more extensible and you won't pay for stuff you don't need in instantiation time.

README.md
```diff
@@ -306,6 +306,39 @@ psReadValues("SELECT ... FROM ...", connection, arg1, arg2);
...
+    }
+)
+```
+
+Or without result callback
+
+```c++
+int myData = 0;
+
+psQuery(
```
OpatrilPeter: I'd prefer to use different name for that, not overload.

MrWarlockX: And what name do you suggest? Something like `psWriteQuery`?

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -131,10 +93,72 @@ namespace SuperiorMySqlpp
...
+     * @tparam warnMode Indicates warning mode level
+     * @tparam ignoreNullable Disables null type checking
+     */
+    template<bool storeResult=detail::PreparedStatementsDefault::getStoreResult(),
+             ValidateMetadataMode validateMode=detail::PreparedStatementsDefault::getValidateMode(),
+             ValidateMetadataMode warnMode=detail::PreparedStatementsDefault::getWarnMode(),
+             bool ignoreNullable=detail::PreparedStatementsDefault::getIgnoreNullable(),
+             typename ResultCallable,
+             typename ParamCallable,
+             typename ConnType>
```
OpatrilPeter, line 195: I don't see a point for templating here.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -131,10 +93,72 @@ namespace SuperiorMySqlpp
...
+    void psQuery(const std::string &query, ConnType &&connection, ParamCallable &&paramsSetter, ResultCallable &&processingFunction)
```
OpatrilPeter: I'd attempt to unify implementations of ps(Read)Query variants if at all possible - the shared implementation would take paramUpdater and resultReader callbacks along with `bool`s that could be used to conditionally execute the callbacks.
Only reason why that wouldn't be a good idea is if compiler wouldn't be able to optimize that away.

tests/db_access/prepared_statements.cpp
```diff
@@ -649,9 +649,20 @@ go_bandit([](){
...
         it("can work with prepared statement helper functions - psReadQuery (invalid data types)", [&](){
-            AssertThrows(PreparedStatementTypeError, psReadQuery("SELECT `id`, `blob`, `binary`, `varbinary` FROM `test_superior_sqlpp`.`binary_data` ORDER BY `id` LIMIT 1", 
+            AssertThrows(PreparedStatementTypeError, psReadQuery("SELECT `id`, `blob`, `binary`, `varbinary` FROM `test_superior_sqlpp`.`binary_data` ORDER BY `id` LIMIT 1",
                 connection, [&](int , int, const BlobData &, const BlobData &) {}));
         });
+
+        it("can work with psQuery helper function", [&](){
+            const int target = 5;
+            int counter = 0;
+            psQuery("SELECT ?", connection,
```
OpatrilPeter: This being very useful function, please consider adding another, more real-world use.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
MrWarlockX: Well, you can call it then like this:
```c++
psQuery("SELECT ...", gimmeConnection(), ....);
```
Or in another words, it gives you ability to use one-time-use connection.

README.md
OpatrilPeter: That implies `INSERT`. I'd suggest to have `psParamQuery` and `psResultQuery` (psReadQuery) as special cases of `psQuery`.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -11,89 +11,127 @@ namespace SuperiorMySqlpp
...
-         * @param query Query for prepared statement
-         * @tparam storeResult Boolean indicating if results will be in `store` or `use` mode
-         * @tparam validateMode Indicates validate mode level
-         * @tparam warnMode Indicates warning mode level
-         * @tparam ignoreNullable Disables null type checking
-         */
-        template <bool storeResult, ValidateMetadataMode validateMode, ValidateMetadataMode warnMode, bool ignoreNullable, typename... Args>
-        auto generatePreparedStatementImpl(Connection &connection, const std::string &query, void (*)(Args...)) {
-            return connection.makePreparedStatement<ResultBindings<std::remove_cv_t<std::remove_reference_t<Args>>...>, storeResult, validateMode, warnMode, ignoreNullable>(query);
+    struct NoCallback
```
OpatrilPeter: I'd call it `EmptyParamCallback`, as this name seem too broad.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
OpatrilPeter: Fair enough.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -11,89 +11,127 @@ namespace SuperiorMySqlpp
...
+             ValidateMetadataMode validateMode=detail::PreparedStatementsDefault::getValidateMode(),
+             ValidateMetadataMode warnMode=detail::PreparedStatementsDefault::getWarnMode(),
+             bool ignoreNullable=detail::PreparedStatementsDefault::getIgnoreNullable(),
+             typename ResultCallable,
+             typename ParamCallable,
+             typename ConnType>
+    void psQuery(const std::string &query, ConnType &&connection, ParamCallable &&paramsSetter, ResultCallable &&processingFunction)
+    {
+        static constexpr bool NoParamCallback = std::is_same<ParamCallable, EmptyParamCallback>::value;
+        static constexpr bool NoResultCallback = std::is_same<ResultCallable, EmptyParamCallback>::value;
```
OpatrilPeter: I called for renaming EmptyCallback to EmptyParamCallback precisely because I didn't expect you to use it for ResultCallback, as their types are logically distinct - EmptyParamCallback returns boolean, EmptyResultCallback does nothing.
That aside, I believe this function will be nicer if you negate the conditions, i.e. call it `hasParamCallback`.
Also, unless you're following some specific formatting style, I wound't use brackets around `break`.

include/superior_mysqlpp/utils.hpp
```diff
@@ -75,4 +75,224 @@ namespace SuperiorMySqlpp
...
+         * @tparam ResultType type returned by function
+         * @tparam IsConst true_type, if function has been marked const
+         * @tparam IsVolatile true_type, if function has been marked volatile
+         * @tparam Args function arguments
+         */
+        template<typename ResultType, uint64_t Flags, typename... Args>
+        struct FunctionInfoImpl
+        {
+            using result_type   = ResultType;
+            using is_const      = std::integral_constant<bool, (Flags & FunctionFlag_IsConst) != 0>;
```
OpatrilPeter, line 104: May I ask why is this preferred over `static constexpr bool`?

include/superior_mysqlpp/utils.hpp
```diff
@@ -75,4 +75,224 @@ namespace SuperiorMySqlpp
...
+    namespace detail
+    {
+        /**
+         * Function information flags
+         */
+        enum : uint64_t
+        {
+            FunctionFlag_IsConst = 1ULL << 0ULL,
+            FunctionFlag_IsVolatile = 1ULL << 1ULL,
+            FunctionFlag_IsLvalueRefCtx = 1ULL << 3ULL,
```
OpatrilPeter: Are you sure this is the correct term? I don't know myself, but in any case, I feel these flags (or appropriate FunctionInfo "fields") deserve comments.
I'd perhaps use the word `this` in it, such as FunctionFlag_ThisIsConst, FunctionFlag_ThisIsLvalueRef etc.

include/superior_mysqlpp/extras/prepared_statement_utils.hpp
```diff
@@ -11,89 +11,127 @@ namespace SuperiorMySqlpp
...
          * @tparam storeResult Boolean indicating if results will be in `store` or `use` mode
          * @tparam validateMode Indicates validate mode level
          * @tparam warnMode Indicates warning mode level
          * @tparam ignoreNullable Disables null type checking
          */
-        template <bool storeResult, ValidateMetadataMode validateMode, ValidateMetadataMode warnMode, bool ignoreNullable, typename CallableClass, typename... Args>
-        auto generatePreparedStatementImpl(Connection &connection, const std::string &query, void (CallableClass::*)(Args...) const) {
-            return connection.makePreparedStatement<ResultBindings<std::remove_cv_t<std::remove_reference_t<Args>>...>, storeResult, validateMode, warnMode, ignoreNullable>(query);
+        template <bool storeResult, ValidateMetadataMode validateMode, ValidateMetadataMode warnMode, bool ignoreNullable, typename ResultCallable, typename ParamCallable>
+        auto generatePreparedStatement(Connection &connection, const std::string &query, ResultCallable &&, ParamCallable &&)
```
OpatrilPeter: Note that in `psQuery`, we use order `paramCallable, resultCallable`, but in internal functions it's switched. I feel that may bite us later, especially as they aren't distinct type-wise.

include/superior_mysqlpp/utils.hpp
MrWarlockX: Because this is rather a trait, so you can easily use it as e.g. overload resolver (having 2 function overloads with one expecting false_type and the other true_type). And value extraction is very simple.

OpatrilPeter: I see it's mostly matter of preference then. Fair enough, I'm ok with this solution.

packages/_debian-common/changelog
```diff
@@ -1,8 +1,12 @@
 libsuperiormysqlpp (0.4.1) UNRELEASED; urgency=medium
 
+  [ Peter Opatril ]
   * refactor: Logging system (DRY)
   * feat: Default logger logs errors
 
+  [ Michal Merc ]
+  * feat: Implement psQuery function
```
OpatrilPeter: I'd mention that you've also added ps{Param|Result}Query. You may of course just mention them as "convenience functions".

README.md
```diff
@@ -261,23 +261,28 @@ while (preparedStatement.fetch())
...
 
 ### Convenience read functions
-**psReadQuery**
+**psParamQuery**
+
+Invokes psQuery with param setter only.
 ```c++
-auto preparedStatement = connection.makePreparedStatement<ResultBindings<Sql::Int, Sql::Int>("SELECT ... FROM ...");
-psReadQuery(preparedStatement, <Callable>);
+psResultQuery("INSERT INTO ... (col1, col2, ...) VALUES (?, ?, ...)", connection, [&](T1 &col1, T2& col2, ...) -> bool {
```
OpatrilPeter: psParamQuery

OpatrilPeter (approved):

PR #114, opened 2019-06-18
refactor: Change order of arguments of psQuery

MrWarlockX:
This PR changes order of arguments of newly-created *psQuery* functions, before version containing them is released.


packages/_debian-common/changelog
```diff
@@ -6,6 +6,7 @@ libsuperiormysqlpp (0.4.1) UNRELEASED; urgency=medium
 
   [ Michal Merc ]
   * feat: Implement psQuery/psParamQuery/psResultQuery convenience functions
```
smradCZ, line 8: Please increase MINOR version, introducing new methods and deprecating another one is not what I expect from patch version change

smradCZ (approved):

PR #115, opened 2019-06-24
feat: add support for debian buster

AdamStepan:


packages/debian-buster/dpkg-jail/debian/control
```diff
@@ -0,0 +1,17 @@
+Source: libsuperiormysqlpp
+Section: unknown
+Priority: extra
+Maintainer: Seznam.cz
+Build-Depends: debhelper,
+               clang,
```
smradCZ: It should be `g++` instead of `clang`

smradCZ (changes requested):

smradCZ (dismissed):

smradCZ (approved):

PR #117, opened 2019-07-15
chore: Update bandit submodule to current HEAD

smradCZ:
Fix header includes and `using namespace snowhouse` because assertion helpers were moved to this namespace.

### DIFF
--- a/include/superior_mysqlpp/extras/row_stream_adapter.hpp
+++ b/include/superior_mysqlpp/extras/row_stream_adapter.hpp
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <cassert>
-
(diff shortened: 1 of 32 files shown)


include/superior_mysqlpp/extras/row_stream_adapter.hpp
```diff
@@ -1,4 +1,7 @@
 #pragma once
+
+#include <cassert>
```
OpatrilPeter: This one seems too paranoid to me (assert(*this);), consider removing it.

OpatrilPeter (dismissed):

OpatrilPeter (approved):

PR #124, opened 2019-09-24
feat: Use converter struct instead of converter function

### LINKED ISSUE #123: Make user-defined converters include-order independent
Currently the users can specify converters for rowstream by template-specializing the `SupericorMySqlpp::Converters::to(const char* str, unsigned int length)` function.

However, this specialization has to be defined before  `superior_mysqlpp` includes, otherwise it is not picked up.

Using templated converter struct instead would not have that problem.

Erbureth:
Fixes #123 with backward compatibility for existing user-defined converters

### DIFF
--- a/include/superior_mysqlpp/converters.hpp
+++ b/include/superior_mysqlpp/converters.hpp
@@ -14,22 +14,58 @@
 
 namespace SuperiorMySqlpp { namespace Converters
 {
-    template<typename T>
-    inline std::enable_if_t<std::is_integral<T>::value, T> to(const char* str, unsigned int length)
-    {
-        return toInteger<T>(str, length);
+    template<typename, typename=void>
+    struct To;
+
+    namespace detail {
+        template<typename... Ts> struct make_void { typedef void type;};
+        template<typename... Ts> using void_t = typename make_void<Ts...>::type;
+
+        template<typename T>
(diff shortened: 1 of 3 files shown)


include/superior_mysqlpp/converters.hpp
```diff
@@ -14,22 +14,45 @@
 
 namespace SuperiorMySqlpp { namespace Converters
 {
-    template<typename T>
-    inline std::enable_if_t<std::is_integral<T>::value, T> to(const char* str, unsigned int length)
-    {
-        return toInteger<T>(str, length);
+    namespace detail {
+        template<typename First, std::size_t>
```
MrWarlockX: It won't work in this way, you need to have that template specialization lookup implemented like so, to ensure proper sfinae.
```c++
        // not needed in c++17
        template<typename... Ts> struct make_void { typedef void type;};
        template<typename... Ts> using void_t = typename make_void<Ts...>::type;

        template<typename T>
        using get_to_method_type = decltype(&To<T>::operator());
        
        template<typename T, typename = void>
        struct has_to_overload : std::false_type {};
        
        template<typename T>
        struct has_to_overload<T, void_t<get_to_method_type<T>>> : std::true_type {};
```

MrWarlockX:
Code LGTM, but I would like to see some comments.

smradCZ (approved):

PR #125, opened 2019-10-23
fix: remove noexcept clause from Statement::close()

tomasprochazka2:


include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -1140,18 +1140,19 @@ namespace SuperiorMySqlpp { namespace LowLevel
             /**
              * Closes the prepared statement and deallocates the statement handler.
              * In case of error, std::terminate() is called.
```
OpatrilPeter: This line is outdated.

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -25,19 +25,26 @@ namespace SuperiorMySqlpp
         using SuperiorMySqlppError::SuperiorMySqlppError;
 
         MysqlInternalError(const std::string& message,
-                           const char* mysqlError,
+                           const char* mysqlError_,
                            unsigned errorCode_)
-            : SuperiorMySqlppError(message + "\nMysql error: " + std::string(mysqlError))
+            : SuperiorMySqlppError(message + "\nMysql error: " + std::string(mysqlError_))
```
OpatrilPeter: optimization: explicit std::string construction is redundant

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -25,19 +25,26 @@ namespace SuperiorMySqlpp
...
+            return mysqlError;
+        }
+
         unsigned getErrorCode() const
         {
             return errorCode;
         }
 
     private:
+        const char* mysqlError;
```
OpatrilPeter: I'm not really comfortable with this. Both Oracle and MariaDB specification of `mysql_error` is not making a strong guarantee about lifetime of the error message beyond "Returns the last error message for the most recent function call that can succeed or fail." That means it's fragile to keep the pointer in a long term storage.
I'd consider just printing the primary message of this exception in `loggerPtr->logMySqlStmtCloseError`.

OpatrilPeter (changes requested):

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -25,19 +26,26 @@ namespace SuperiorMySqlpp
         using SuperiorMySqlppError::SuperiorMySqlppError;
 
         MysqlInternalError(const std::string& message,
-                           const char* mysqlError,
+                           const std::string& mysqlError_,
```
OpatrilPeter: `const char *` here is ok, you're making temporary dynamic string otherwise.

include/superior_mysqlpp/exceptions.hpp
```diff
@@ -25,19 +26,26 @@ namespace SuperiorMySqlpp
...
+            return mysqlError;
+        }
+
         unsigned getErrorCode() const
         {
             return errorCode;
         }
 
     private:
+        const std::string mysqlError;
```
OpatrilPeter: I'd remove const here and on member below, it seems like a style antipattern without clear purpose.

OpatrilPeter (changes requested):
Mentioned details aside, looks good to me.

OpatrilPeter (approved):

PR #126, opened 2019-10-31
Release draft with artifacts on tag

smradCZ:
+ fixed intermittently failing test on connection pool
+ fixed populationId in connection pool

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,45 @@ language: cpp
 os: linux
 sudo: required
 services: docker
+
 branches:
   only:
-    - master
-    - devel
+  - master
+  - devel
+  # regex to match valid tag format or TravisCI won't trigger deploys when a new tag is pushed
+  - /^v[0-9]+\.[0-9]+\.[0-9]+.*$/
+
 script:
-  - make CONCURRENCY=4 package-debian-buster-dbuild
-  - make CONCURRENCY=4 package-debian-stretch-dbuild
-  - make CONCURRENCY=4 package-debian-jessie-dbuild
-  - make CONCURRENCY=4 package-fedora-22-dbuild
+- make CONCURRENCY=4 package-debian-buster-dbuild
+- make CONCURRENCY=4 package-debian-stretch-dbuild
+- make CONCURRENCY=4 package-debian-jessie-dbuild
+- make CONCURRENCY=4 package-fedora-22-dbuild
+
+before_deploy: |
+  # move packages to ARTIFACTS directory
+  pushd packages
+  mkdir ARTIFACTS
+  ## rename debian packages to be distribution specific and move them to ARTIFACTS directory
+  for DEBIAN_PKG in debian-*/libsuperiormysqlpp-dev_*.deb; do
+    GITHUB_ARTIFACT_NAME=`echo ${DEBIAN_PKG} | sed --regexp-extended 's|debian-([^/]+)/libsuperiormysqlpp-dev_([^_]+)_amd64.deb|ARTIFACTS/libsuperiormysqlpp-dev_\2-\1_amd64.deb|'`
+    echo "Renaming & moving package '${DEBIAN_PKG}' to '${GITHUB_ARTIFACT_NAME}'"
+    mv ${DEBIAN_PKG} ${GITHUB_ARTIFACT_NAME}
+  done
+  ## move RPM packages to ARTIFACTS directory
+  mv --target-directory=ARTIFACTS fedora-*/*.rpm
+  popd
+
+  # GitHub releaser script (https://github.com/lindell/github-release-cli)
+  # used because standard github-release deploy module doesn't support specifying release message from ENV
+  curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis --fail --location --output github-releaser && chmod +x github-releaser \
+    || (echo "Failed to download github-releaser script" && return 1)
+  ## get changes from debian's changelog
+  export BODY=`dpkg-parsechangelog --file packages/_debian-common/changelog --show-field Changes | grep --invert-match --extended-regexp '^$|^libsuperiormysqlpp |^  \[ '`
+  export FILES=packages/ARTIFACTS/*
+
+deploy:
+  provider: script
+  script: ./github-releaser -draft -verbose
+  skip_cleanup: true
+  on:
+    tags: true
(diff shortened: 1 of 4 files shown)


tests/db_access/connection_pool.cpp
```diff
@@ -90,8 +89,10 @@ go_bandit([](){
                 AssertThat(poolState.available, Equals(0U));
 
                 connectionPool.get();
-
-                std::this_thread::sleep_for(2s);
+                backoffSleep(2s, [&](){
+                    auto&& poolState = connectionPool.poolState();
```
OpatrilPeter: I'd refactor this pattern into function, this test is full of needless boilerplate.
```
assertPoolState = [&connectionPool](size, available) {
            auto&& poolState = connectionPool.poolState();
            AssertThat(poolState.size, Equals(size));
            AssertThat(poolState.available, Equals(available));
};
```
Of course, it doesn't cover this specific (nonasserting, boolean returning) pattern, so second similar function checkPoolState ought to be made ... but if the last test truly needs >= condition, perhaps it's not worth trying to generalize it.

.travis.yml
```diff
@@ -2,12 +2,40 @@ language: cpp
...
+- make CONCURRENCY=4 package-debian-buster-dbuild
+- make CONCURRENCY=4 package-debian-stretch-dbuild
+- make CONCURRENCY=4 package-debian-jessie-dbuild
+- make CONCURRENCY=4 package-fedora-22-dbuild
+
+before_deploy: |
+  # rename debian packages to be distribution specific
+  pushd packages
+  for DEBIAN_PKG in `ls debian-*/libsuperiormysqlpp-dev_*.deb`; do
+    GITHUB_ARTIFACT_NAME=`echo ${DEBIAN_PKG} | sed 's|debian-\([^/]*\)/libsuperiormysqlpp-dev_\([^_]\+\)_amd64.deb|debian-\1/libsuperiormysqlpp-dev_\2-\1_amd64.deb|'`
```
OpatrilPeter: `debian-\([^/]*\)/` should be `debian-\([^/]\+\)/`, I believe.
I'd still consider following for clarity.
```
sed -E 's|debian-([^/]+)/libsuperiormysqlpp-dev_([^_]+)_amd64.deb|debian-\1/libsuperiormysqlpp-dev_\2-\1_amd64.deb|'
```

.travis.yml
```diff
@@ -2,12 +2,40 @@ language: cpp
...
-  - make CONCURRENCY=4 package-fedora-22-dbuild
+- make CONCURRENCY=4 package-debian-buster-dbuild
+- make CONCURRENCY=4 package-debian-stretch-dbuild
+- make CONCURRENCY=4 package-debian-jessie-dbuild
+- make CONCURRENCY=4 package-fedora-22-dbuild
+
+before_deploy: |
+  # rename debian packages to be distribution specific
+  pushd packages
+  for DEBIAN_PKG in `ls debian-*/libsuperiormysqlpp-dev_*.deb`; do
```
OpatrilPeter: I recall `ls` not being recommended for noninteractive/scripting uses, mostly for output ambiguity, you can see the discussion on the topic for example here - https://unix.stackexchange.com/questions/128985/why-not-parse-ls-and-what-to-do-instead.
That said, we're working with finite directory tree with no weird cases, so may be fine.

OpatrilPeter: also, I don't think you really need ls here, just globbing is likely enough -
```bash
for DEBIAN_PKG in debian-*/libsuperiormysqlpp-dev_*.deb; do
```
consider `shopt -s nullglob`

.travis.yml
```diff
@@ -2,12 +2,40 @@ language: cpp
...
+    mv ${DEBIAN_PKG} ${GITHUB_ARTIFACT_NAME}
+  done
+  popd
+  # GitHub relaser script (https://github.com/lindell/github-release-cli)
+  # used because standard github-release deploy module doesn't support specifying release message from ENV
+  curl https://github.com/lindell/github-release-cli/releases/download/1.3.0/github-releaser-travis --fail --location --output github-releaser && chmod +x github-releaser \
+    || (echo "Failed to download github-releaser script" && return 1)
+  # get changes from debian's changelog
+  export BODY=`dpkg-parsechangelog --file packages/_debian-common/changelog --show-field Changes | grep --invert-match --extended-regexp '^$|^libsuperiormysqlpp |^  \[ '`
+  export FILES=`ls packages/*/libsuperiormysqlpp[_-]*.@(deb|rpm)`
```
OpatrilPeter: how about moving the packages to a special directory and then set FILES just to the `*` of dir? It would certainly be more straightforward.

smradCZ: OK, packages are moved to special directory, so globbing pattern is simplified now.

.travis.yml
smradCZ: Good point, thanks.

tests/db_access/connection_pool.cpp
```diff
@@ -34,67 +33,55 @@ go_bandit([](){
     describe("Test connection pool", [&](){
         auto& s = getSettingsRef();
 
+
         it("can clear pool properly", [&](){
             auto&& connectionPool = makeConnectionPool<true, true, true, true, true>([&](){
                 return std::async(std::launch::async, [&](){ return std::make_shared<Connection>(s.database, s.user, s.password, s.host, s.port); });
             });
+            auto waitForPoolState = [&connectionPool](std::chrono::milliseconds max, auto size, auto available) {
```
OpatrilPeter: In this case, I'd prefer `std::size_t` over `auto`.

OpatrilPeter: Also, this implementation of these conveniency functions is totally ok, but considering the fact that you use them in multiple tests, it may be better overall to define them as conventional, standalone static functions with explicit connectionPool parameter.

tests/db_access/connection_pool.cpp
```diff
@@ -21,80 +21,72 @@ using namespace SuperiorMySqlpp;
...
 using namespace std::chrono_literals;
 
-
 auto makeSharedPtrConnection()
 {
     auto& s = getSettingsRef();
     return std::async(std::launch::async, [&](){ return std::make_shared<Connection>(s.database, s.user, s.password, s.host, s.port); });
 }
 
+template <typename T>
```
OpatrilPeter: I'd leave private helper functions in anonymous namespace.

OpatrilPeter (approved):

PR #128, opened 2020-08-03
Fix string to float conversion

OpatrilPeter:
Additionally solves other minor issues.

Note several breaking changes.

### DIFF
--- a/include/superior_mysqlpp/types/array.hpp
+++ b/include/superior_mysqlpp/types/array.hpp
@@ -14,18 +14,32 @@ namespace SuperiorMySqlpp
 {
     /**
      * Base class for all array based SQL types
+     *
+     * If TerminatingZero == true, hidden empty byte is always appended after payload.
+     * Mind the fact that storage needs to be (at least one byte) larger for this mode.
      */
-    template<std::size_t N>
+    template<std::size_t N, bool TerminatingZero = false>
     class ArrayBase
     {
     protected:
-        std::array<char, N> array;
+        std::array<char, N + int(TerminatingZero)> array;
 
-        using ItemsCount_t = unsigned long;
+        using ItemsCount_t = std::size_t;
         ItemsCount_t itemsCount = 0;
 
     public:
-        ArrayBase() = default;
+        /**
+         * Note: should be constexpr, sadly GCC 4.9.2 from Debian Jessie doesn't implement C++14 sufficiently
+         */
+        ArrayBase()
+        {
+            if (TerminatingZero)
+            {
+                array.front() = '\0';
+                // back() is just redundant sanity check
+                array.back() = '\0';
+            }
+        }
 
         ArrayBase(const ArrayBase& other)
         {
@@ -62,6 +76,10 @@ namespace SuperiorMySqlpp
         {
             std::memcpy(array.data(), other.array.data(), other.size());
             itemsCount = other.size();
+            if (TerminatingZero)
+            {
+                *this->end() = '\0';
+            }
         }
 
         template<std::size_t NN = N>
@@ -73,7 +91,7 @@ namespace SuperiorMySqlpp
             return itemsCount;
         }
 
-        const ItemsCount_t& counterRef() const
+        constexpr const ItemsCount_t& counterRef() const
         {
             return itemsCount;
         }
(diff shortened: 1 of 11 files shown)


include/superior_mysqlpp/types/array.hpp
```diff
@@ -14,18 +14,29 @@ namespace SuperiorMySqlpp
 {
     /**
      * Base class for all array based SQL types
+     *
+     * If TerminatingZero == true, hidden empty byte is always appended after payload.
+     * Mind the fact that storage needs to be (at least one byte) larger for this mode.
      */
-    template<std::size_t N>
+    template<std::size_t N, bool TerminatingZero = false>
     class ArrayBase
```
MrWarlockX, line 22: Just tip for the future - maybe it would be worth it to change this type to `template<typename, size_t, bool>` and possibly create an alias e.g. `template<size_t sz, boo; b> using ByteArrayBase = ArrayBase<(unsigned?) char, sz, b>;`.
The reason why I suggest this is simple: alignment and reusability.

include/superior_mysqlpp/types/blob_data.hpp
```diff
@@ -17,15 +17,19 @@
 
 namespace SuperiorMySqlpp
 {
-    template<std::size_t N>
-    class BlobDataBase : public ArrayBase<N>
+    template<std::size_t N, bool TerminatingZero = false>
+    class BlobDataBase : public ArrayBase<N, TerminatingZero>
```
MrWarlockX, line 21: Similarly as in comment above (ArrayBase). User can have blob data of e.g. type integer or some structure. This way they have to copy data via e.g. `memcpy` to aligned storage to ensure they don't have problems with aliasing.

include/superior_mysqlpp/types/array.hpp
```diff
@@ -170,7 +197,7 @@ namespace SuperiorMySqlpp
 
         constexpr auto maxSize() const
         {
-            return array.max_size();
+            return array.max_size() - int(TerminatingZero);
```
MrWarlockX: What about this way?
```suggestion
            return N;
```

MrWarlockX (dismissed):
I didn't know about the missing template arguments pass in prepared statement utils. Good thing that it is fixed.

include/superior_mysqlpp/types/array.hpp
```diff
@@ -135,22 +150,34 @@ namespace SuperiorMySqlpp
 
         auto& front()
         {
-            return array.front();
+            if (count() == 0)
```
PredatorCZ:
```suggestion
            if (empty())
```

include/superior_mysqlpp/types/array.hpp
```diff
@@ -170,7 +197,7 @@ namespace SuperiorMySqlpp
 
         constexpr auto maxSize() const
         {
-            return array.max_size();
+            return array.max_size() - int(TerminatingZero);
         }
 
         bool empty() const
```
PredatorCZ, line 206: I might ask a lot, but can you add
```suggestion
        constexpr bool empty() const
```
here?

OpatrilPeter: Many other methods are constexpr already, so why not.

However, it's not that useful for two reasons:
- ideally, it should be added everywhere consistently to be useful, notably ArrayBase constructors.  If you'll do this you'll notice that Debian Jessie version of GCC has imperfect support for C++14 constexpr feature extensions and will reject it
- secondly, one has to wonder if there is any practical usage for this particular type in compile-time at all, as it's designed for database I/O

include/superior_mysqlpp/types/array.hpp
OpatrilPeter: I agree that this idea has merit, but I'd rather pursue it later, in more mature and robust form that allows users to deserialize binary data to arbitrary types - solving problems like alignment and `std::string` support in one go.
In any case, I feel it's out of scope of this PR.

include/superior_mysqlpp/types/array.hpp
PredatorCZ: I just assumed, since it's already in `count()` method, it shouldn't hurt here as well. ^_^

include/superior_mysqlpp/types/array.hpp
MrWarlockX: I agree.

MrWarlockX (approved):
Looks good to me.

PR #131, opened 2020-12-17
Multiple definition of mapSecondGetter function

daniele77:
An `inline` is missed in the global lambda function `mapSecondGetter`. 
Without it, a linker error is produced when the header file `superior_mysqlpp.hpp` is included in more than one translation unit

### DIFF
--- a/include/superior_mysqlpp/master_slave_connection_pools.hpp
+++ b/include/superior_mysqlpp/master_slave_connection_pools.hpp
@@ -17,8 +17,9 @@
 
 namespace SuperiorMySqlpp
 {
-    inline auto mapSecondGetter = [](auto&& item) -> auto& { return std::forward<decltype(item)>(item).second; };
-
+    template <typename T>
+    T& mapSecondGetter(T&& item) { return std::forward<T>(item).second; }
+    
     template<
         typename SharedPtrPoolType,
         typename SlaveId=unsigned int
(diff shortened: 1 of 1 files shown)


daniele77:
On older compilers, the `inline` directive does not work on `auto` lambdas.
This is the right code, instead:

    template <typename T>
    T& mapSecondGetter(T&& item) { return std::forward<T>(item).second; }

     // auto mapSecondGetter = [](auto&& item) -> auto& { return std::forward<decltype(item)>(item).second; };

daniele77:
closes #131

daniele77:
close #132

OpatrilPeter:
Thank you for your interest and bringing it to our attention.

> template <typename T>
T& mapSecondGetter(T&& item) { return std::forward<T>(item).second; }

Firstly, that doesn't look right to me - item and result types are independent - in fact, T is pair while result type is second element. Following should probably work in C++14.
```cpp
template <typename T>
auto& mapSecondGetter(T&& item) { return std::forward<T>(item).second; }
```

Secondly, lambda with auto in parameter ought to be templated implicitly, isn't the problem actually with the "global variable"? Using `const auto` instead of `auto` on the lambda should fix the problem, as global constants are implicitly `static`. I don't have GCC 10 at hand, so I can't test - if adding const won't help, posting the error message would be helpful.

daniele77:
Thank you.
Using `const auto` fixed the issue (with GCC 10).
I updated the PR.
Do you think the fix will be available in the next release?

OpatrilPeter:
Thank you for your patience. I'll merge the changes and release a new version soon (note that we have recently discontinued OS-specific releases (such as deb packages), as they made little sense for header-only library).

As we've also changed our policy on commit messages, I went ahead and recreated the changes in different PR, so I'll be closing this one.

PR #133, opened 2021-01-11
set LD_LIBRARY_PATH in runtest.sh

KristianMika:
Running tests on Debian Stretch with Clang-5.0 fails due to inability to find the `libclang_rt.asan-x86_64.so` file. 
Ends with this error: `error while loading shared libraries: libclang_rt.asan-x86_64.so: cannot open shared object file: No such file or directory`.

My proposed fix sets the `LD_LIBRARY_PATH` variable in the `runtest.sh` script.


KristianMika:
Update: The error occurred due to incorrect configuration of my test environment, therefore I am closing this PR.

PR #137, opened 2022-03-03
feat: add passing client flags into options

PredatorCZ:
Implemented passing client flags into Connection.

### DIFF
--- a/include/superior_mysqlpp/low_level/dbdriver.hpp
+++ b/include/superior_mysqlpp/low_level/dbdriver.hpp
@@ -105,6 +105,8 @@ namespace SuperiorMySqlpp { namespace LowLevel
         /** Logger instance pointer. */
         Loggers::SharedPointer_t loggerPtr;
 
+        unsigned long clientFlags = 0;
+
         using size_t = unsigned long long;
 
     private:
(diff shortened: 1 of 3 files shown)


include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -404,6 +406,41 @@ namespace SuperiorMySqlpp { namespace LowLevel
             return mysql_character_set_name(getMysqlPtr());
         }
 
+        enum class ClientFlags : unsigned long
+        {
+            longPassword = CLIENT_LONG_PASSWORD,
```
OpatrilPeter, line 411: style: Enum values should use PascalCase, not camelCase (I know we're inconsistent in some cases)

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -404,6 +406,41 @@ namespace SuperiorMySqlpp { namespace LowLevel
...
+            rememberOptions = CLIENT_REMEMBER_OPTIONS,
+            pluginAuth = CLIENT_PLUGIN_AUTH,
+            connectAttrs = CLIENT_CONNECT_ATTRS,
+            pluginAuthLenEncClientData = CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA,
+            canHandleExpiredPasswords = CLIENT_CAN_HANDLE_EXPIRED_PASSWORDS,
+            sessionTrack = CLIENT_SESSION_TRACK,
+            deprecateEOF = CLIENT_DEPRECATE_EOF,
+        };
+
+        void addClientFlag(ClientFlags flag) noexcept
```
OpatrilPeter, line 438: rather than this, I'd consider treat it as bitfield and support combination like so
```
A operator| (A a, A b) {
    using T = std::underlying_type_t<A>;
    return A(T(a) | T(b));
}

A& operator|= (A& self, A other) {
    return self = self | other;
}
```
and while we're at it, it might be better here to use conventional `enum` over `enum class`, as enum class doesn't support unenumerated combinations. This way we'd get the piping operators for free too

include/superior_mysqlpp/low_level/dbdriver.hpp
```diff
@@ -105,6 +105,8 @@ namespace SuperiorMySqlpp { namespace LowLevel
         /** Logger instance pointer. */
         Loggers::SharedPointer_t loggerPtr;
 
+        unsigned long clientFlags = 0;
```
OpatrilPeter, line 108: I think we should aim for a solution that doesn't require storing data here that are required only during initialization.

include/superior_mysqlpp/low_level/dbdriver.hpp
OpatrilPeter: On second thought, I prefer the fact it shares the argument of Connection ctor and that requires strong typing.

OpatrilPeter:
We've iterated on this proposal and went with alternative https://github.com/seznam/SuperiorMySqlpp/pull/138 adressing key points
- values are opaque integers reflecting the fact that different backing C libraries may support different flags (or new flags in the future) and conceptually, this is "escape hatch" style of tuning anyway, where we do not guarantee that given flags are well respected by our wrapping library
- we do not store additional field in connection

PR #144, opened 2024-06-19
update github actions and docker image

PredatorCZ:


.github/workflows/cmake.yml
```diff
@@ -1,16 +1,20 @@
...
+  schedule:
+    - cron: '0 12 */7 * *'
 
 env:
   BUILD_TYPE: Debug
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
smradCZ: I would prefer to use the latest LTS version 24.04 of Ubuntu or use simply the `ubuntu-latest` - in that case we get always latest distro for build. When something fail, we will know because we have scheduled builds already.

smradCZ (approved):

